PR #8, opened 2017-07-18
Definition of Done

mmunozba:
The definition of done still requires reviewing.
A quick refresher on what a DoD is: 
https://www.leadingagile.com/2017/02/definition-of-done/

### DIFF
--- a/docs/dev/dod.md
+++ b/docs/dev/dod.md
@@ -1,0 +1,16 @@
+# Definition of Done
+These criteria must be met before a feature or artifact can be considered "done".
+
+## Code
+- Code coverage: >75%
+- Codacy grade A
+- Travis Build runs without errors (green)
+- Meaningful javadoc comments for:
+  - All classes
+  - All non-private methods and fields
+  - Private methods with non-obvious implementation
+
+## General
+- Documentation was updated to reflect the current release candidate
+- Documentation contains no spelling mistakes or grammatical errors
+- Pull request was reviewed by atleast two independent reviewers
--- a/docs/dev/dod.md
+++ b/docs/dev/dod.md
@@ -5,12 +5,10 @@ These criteria must be met before a feature or artifact can be considered "done"
 - Code coverage: >75%
 - Codacy grade A
 - Travis Build runs without errors (green)
-- Meaningful javadoc comments for:
-  - All classes
-  - All non-private methods and fields
-  - Private methods with non-obvious implementation
+- Meaningful javadoc comments for all classes, methods and fields
+  - Can be omitted for trivial methods, such as get() and set()
 
 ## General
 - Documentation was updated to reflect the current release candidate
 - Documentation contains no spelling mistakes or grammatical errors
-- Pull request was reviewed by atleast two independent reviewers
+- Pull request was reviewed by at least two independent reviewers
--- a/docs/dev/dod.md
+++ b/docs/dev/dod.md
@@ -4,7 +4,7 @@ These criteria must be met before a feature or artifact can be considered "done"
 ## Code
 - Code coverage: >75%
 - Codacy grade A
-- Travis Build runs without errors (green)
+- Travis builds the project and runs all tests without error
 - Meaningful javadoc comments for all classes, methods and fields
   - Can be omitted for trivial methods, such as get() and set()
 


docs/dev/dod.md
```diff
@@ -0,0 +1,16 @@
+# Definition of Done
+These criteria must be met before a feature or artifact can be considered "done".
+
+## Code
+- Code coverage: >75%
```
mklopp: reasonable

docs/dev/dod.md
```diff
@@ -0,0 +1,16 @@
...
+- Travis Build runs without errors (green)
+- Meaningful javadoc comments for:
+  - All classes
+  - All non-private methods and fields
+  - Private methods with non-obvious implementation
+
+## General
+- Documentation was updated to reflect the current release candidate
+- Documentation contains no spelling mistakes or grammatical errors
+- Pull request was reviewed by atleast two independent reviewers
```
nfode: atleast -> at least

nfode: brought that typo in by myself 😂

docs/dev/dod.md
```diff
@@ -0,0 +1,16 @@
+# Definition of Done
+These criteria must be met before a feature or artifact can be considered "done".
+
+## Code
+- Code coverage: >75%
+- Codacy grade A
```
mklopp: maybe to much expected because codacy may have some weird expecations ( as seen in the "vorprojekt")

docs/dev/dod.md
```diff
@@ -0,0 +1,16 @@
...
+- Travis Build runs without errors (green)
```
mklopp: elementary, maybe state that travis build means all tests are running

docs/dev/dod.md
```diff
@@ -0,0 +1,16 @@
...
+- Meaningful javadoc comments for:
+  - All classes
+  - All non-private methods and fields
+  - Private methods with non-obvious implementation
```
mklopp: May just request javadoc for all methods, either private or non private, to make things linear

docs/dev/dod.md
mklopp: Makes sense

docs/dev/dod.md
nfode: Yeah you are right but you can disable the "weird excpecations". If some codacy issues cannot be solved or is not needed in the eyes of the team we can disable this check.

docs/dev/dod.md
hnicke: We should add that for trivial methods (like getters and setters) javadocs can be omitted. I rather have no javadoc than something like getX(): 'returns the x'.
Regarding 'Private methods with non-obvious implementation': Explanation of implemention details do not belong into javadocs. they have to be written as normal comments inside the methods. Remember, javadoc is mainly for the consumer of the code.

mmunozba: > Private methods with non-obvious implementation

This line was not intended to request implementation details in javadoc comments, but rather to allow omitting of javadoc comments for methods that do not require them such as get() and set() methods, just like @hnicke described. It makes sense to just allow omitting javadoc for get/set etc. for all methods and not just for private methods like it was initially described. So it would probably be better to just request javadoc for all methods, regardless of visibility, like suggested by @mklopp.

docs/dev/dod.md
mmunozba: I think we should keep grade A for now, if it shows to be too much effort, the document can still be adjusted. Right now it seems manageable.

docs/dev/dod.md
mmunozba: I need a little help here, how would you phrase this?

nfode: > Travis builds the project and runs all tests without error.

nfode: *errors (GitHub on the phone sucks)

mklopp: Agree

mmunozba:
@hnicke raised the question whether javadoc should really be required for all fields. 

Currently, we require a javadoc comment for every field regardless of visibility or any other factors.
So, the question would is: **Which fields should require javadoc comments and which should not?**

For example, some developers just add javadoc comments for non-private fields.
While this would lessen the documentation effort, it would also mean that the documentation is no longer uniform for the whole code base.

I'd like to hear some more opinions and ideas on this issue.

docs/dev/dod.md
c-mueller: Its also possible to customize the criterias, which Codacy will check. By default some Problems are allowed to get grade A if the project is big enough.

docs/dev/dod.md
```diff
@@ -0,0 +1,14 @@
+# Definition of Done
+These criteria must be met before a feature or artifact can be considered "done".
+
+## Code
+- Code coverage: >75%
+- Codacy grade A
+- Travis builds the project and runs all tests without error
```
c-mueller: As discussed in the meeting on sunday, this might need some exceptions. Because some tests might not be able to be performed in the 50 minute timeframe that travis provides.

A external CI server like Jenkins might be a solution for this

koppor: Maybe, you should distinguish between module tests and integration tests. Normal software projects, don't run in timeout issues on travis.

Further, matrix builds can help: https://docs.travis-ci.com/user/speeding-up-the-build/. One can do tests in parallel.

mmunozba:
I think the DoD in its current form is sufficient for our first or first few sprints. 

How testing and builds with Travis will work can be specified further when we set them up. 
Other points such as the required code coverage might also need further change but that will become apparent during later project stages. For now, these requirements will stay as they are.

I propose that we use this DoD for now and schedule another set of reviews after we progressed further with the project.

PR #20, opened 2017-07-27
[WIP] REST API draft

nfode:
Pull Request to review the REST API draft.


docs/draft/rest_api.md
```diff
@@ -0,0 +1,34 @@
...
+  ```
+  GET status/
+  ```
+- the result of the transformation: file/archive
+  ```
+  GET transformation/{id}
+  ```
+- status of the current transformation
+  ```
+  GET transformation/status/{id}
```
c-mueller: I think it would be better to put the id parameter in the middle:
`transformation/{id}/status`

docs/draft/rest_api.md
```diff
@@ -0,0 +1,34 @@
...
+  ```
+- provide log files for a specific transformation
+  ```
+  GET transformation/log/{id}
```
c-mueller: I think it would be better to put the id parameter in the middle:
`transformation/{id}/log`

docs/draft/rest_api.md
```diff
@@ -0,0 +1,34 @@
+# First Draft for the REST API
```
c-mueller: This comment is about the whole Document. This might because of it beeing a very eary draft. What do the API Calls return? JSON?

docs/draft/rest_api.md
```diff
@@ -0,0 +1,34 @@
...
+- status of the current transformation
+  ```
+  GET transformation/status/{id}
+  ```
+- provide log files for a specific transformation
+  ```
+  GET transformation/log/{id}
+  ```
+
+## Data the REST API should be able to receive [POST]
```
c-mueller: As a Response for the POST-Request the id has to be returned.

docs/draft/rest_api.md
```diff
@@ -0,0 +1,34 @@
+# First Draft for the REST API
+> This is only the first draft. Some adaptations are still possible.
+## Data that needs to be provided by the API [GET]
+- current status of the transformer
+  ```
+  GET status/
```
c-mueller: It would be quite useful to be able to query all currently running Transformations. Maybe the status request should return those.

docs/draft/rest_api.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Data the REST API should be able to receive [POST]
+- A single call containing a CSAR - archive and the target platform
+  (for some platforms some credentials are also necessary)
+  ```
+  POST transformation/
+  ```
+
+## How to process a lot of incoming calls at once?
+- 1. Option: Create a queue, they then get transformed one by one
```
c-mueller: This is probably the better solution for now.

docs/draft/rest_api.md
nfode: Yeah forgot this one. JSON is the best solution for all kind of applications.

docs/draft/rest_api.md
nfode: 👍

docs/draft/rest_api.md
nfode: 👍

docs/draft/rest_api.md
nfode: 👍

docs/draft/rest_api.md
```diff
@@ -0,0 +1,111 @@
...
+    "id": 1,
+    "status": "inqueue|started|stopped|failed",
+    "progress":  "1-100"
+  }
+  ```
+  *status*: returns the current status
+  *progress*: returns the progress as percentage from 1-100%
+- provide log files for a specific transformation
+  ```
+  GET transformation/{id}/log
```
c-mueller: Maybe we should discuss the behaviour of this. I think putting the whole log into one Sting is not the right solution. I think about the following changes/additons:

- replace the one string with a Array of strings. Each element represents one line in the log.
- because those logs can get very long we should implement some mechanism that limits the linecount by default. For example: If you call the request (with no extra parameter) you should receive the most recent 50 lines of the log. If you want more or everything a parameter has to be added.

docs/draft/rest_api.md
```diff
@@ -0,0 +1,111 @@
...
+- status of the current transformation
+  ```
+  GET transformation/{id}/status
+  ```
+  **on success returns:**
+  ```json
+  {
+    "id": 1,
+    "status": "inqueue|started|stopped|failed",
+    "progress":  "1-100"
```
c-mueller: Just giving the progress value seems okay in combination with the logs. However i would still add a message field. this message field jst sends the current stage of the deployment. Like: Parsing, Generating Docker Images (For kubernetes)...

This also results in a question to the customer: Localisation of the application. If we just use english everywhere there is no need for localisation. however if we have to localize the frontend application we can't just send a text message.

This functionality could possibly be realized using the status field.

docs/draft/rest_api.md
```diff
@@ -0,0 +1,111 @@
...
+All calls and requests return data as JSON.
+## Data that needs to be provided by the API
+- current status of the transformer
+  ```
+  GET status/
+  ```
+  **returns:**
+  ```json
+  {
+    "state": "error|transformation|idle",
```
c-mueller: the state name`transformation` should be renamed to `transforming` or `running`

docs/draft/rest_api.md
```diff
@@ -0,0 +1,111 @@
...
+    "current_transformations": [1,2,3]
+  }
+  ```
+  *state*: current status of the transformer
+  *current_transformations:* ids of the active or queued transformations
+- the result of the transformation: file/archive
+  ```
+  GET transformation/{id}
+  ```
+  **on success returns:**
```
c-mueller: This has to be redone. If i understood this correct  this should return the result of the transformation once its done. Something like this cannot be done in a string. Lets say the resulting archiveis 100 MB large. then we should not put the data into a json file. Maybe file points to some HTTP path with wich the user can download the archive.

c-mueller: Another thing here: What happens if the transformation is not done yet?

c-mueller:
One thing the API currently does not cover is additionally required user input. I had the following idea for this.

If extra user input is required the status of the transformation will change to `user-input` (or similar).
With this status change the frontend application has to call a specific GET request that returns the questions the user has to answer. I was thinking of the following Path `GET /transformation/{id}/questions` the response will look like this:
```json
{
  "questions": [
    {
      "id": 1,
      "question": "Please enter the API token for AWS:",
      "inputType": "string"
    },
    {
      "id": 2,
      "question": "Please enter the password for the MySQL root account:",
      "inputType": "password"
    }
  ]
}
```
To quickly explain this: the questions array contains a list of question objects. Each question object has a identifier (maybe even a string), the text for the question and the input type. The reason for giving the input type might be useful to not display passwords on screen just like regular text.

In order to answer the Questions the client has to POST on the same URL with a Data body looking like this:
```json
{
  "responses": [
    {
      "id": 1,
      "answer": "API TOKEN"
    },
    {
      "id": 2,
      "answer": "MYPASS"
    }
  ]
}
```

nfode:
Maybe I can fix that stuff this evening if not or if you have some additions just add it to the document. @c-mueller semms to know what he does.

c-mueller:
I will try to add some things later. Gonna have to work on the architiceture draft for a bit though too.

And yes its true i have worked with stuff like this before.

docs/draft/rest_api.md
nfode: I would use transforming.

docs/draft/rest_api.md
nfode: Returns a error.

docs/draft/rest_api.md
nfode: Yeah that would be nice!

docs/draft/rest_api.md
nfode: Also a good idea!

nfode:
At the first look the idea with the questions in not bad but i would add that passwords should be encrypted.

docs/draft/rest_api.md
nfode: Isnt it possible to convert a file to a binary blob and send it as string?

c-mueller:
Encrypting passwords is not easy because you cannot use symetric encryption like AES. For this to work both parties need to know the key and this key has to be sent to the other party somehow. If this transmission happens through a unencrypted channel the whole encryption is useless.

Using Asymetric Crypto like RSA might be a idea however it seems overcomplicate the whole thing by alot. The best way to encrypt the passwords and anything else is to just use HTTPS instead of HTTP for transmission. As far as I know most (if not every) web application transmitts the password or its hash in plaintext that gets encrypted by the HTTPS protocoll

nfode:
Closed in favour of the alternative draft. See PR #29

PR #29, opened 2017-08-16
Rest API draft

hnicke:
There are open issues..see the TODO list in the changelog.md file


hnicke:
I decided to drop the `ID` fiield  of the resource `toscamodel`. Identifying field is now the `appName` field. 
Reasons:
- While thinking about how we store the artifacts the best way, i thought about ditching relationale databases and storing everything on disk, because we don't have indexable information*, only binary data. I'd like to store our data on the filesystem, while using the rest api structure also as the directory structure.
- the directory structure is better human readable with 'appName' as a folder name instead of 'ID'

*While we have indexable data during transformation, it can be derived almost completely completely from the CSAR, so It's not worth storing it.  Only key-value pairs (user-input) maybe needs to be stored, this should be done with simple .property files.

docs/draft/rest_api.md
```diff
@@ -1,201 +1,324 @@
...
-  
-HTTP Response Code: 200
+*returns:*
 ```json
-{
-  "id": 1,
-  "result": "/results/transform_1.zip"
-}
+[ 
+    "aws": {
```
nfode: missing opening curly braces

docs/draft/rest_api.md
```diff
@@ -1,201 +1,324 @@
...
+        "href":"/platforms/aws"
+    },
+    "cloudformation": {
+        "href":"/platforms/cloudformation"
+    }
```
nfode: missing closing braces

docs/draft/rest_api.md
```diff
@@ -1,201 +1,324 @@
...
-*Parameters*:
-- **id**: The id of the transformation for which the result should be retrieved.
+*returns*:
+```json
+[
+    {
+        "href":"/toscamodels/hello-world",
+        "name":"hello-world",
+        "csar": {
+            "href":"/toscamodels/hello-world/csar"
```
nfode: Not sure if it is necessary to return the href. It might be easier to use but is in my opinion unnecessary extra information. You can build this href with the name.

But maybe I skipped a information and this common practice.

docs/draft/rest_api.md
nfode: this is not really important because this is only a draft and no specification

docs/draft/rest_api.md
hnicke: When the supervisors talked to us about REST, they emphasized self discovery of REST services. That means: The clients should not have to build its own uri's, but rather extract the full uris from the response. I did some research and this is indeed considered best practice.
But I'm aware that I did not embed this practice consistently. We should work on that!

docs/draft/rest_api.md
```diff
@@ -1,201 +1,293 @@
...
-  
-HTTP Response Code: 200
+*returns:*
 ```json
-{
-  "id": 1,
-  "result": "/results/transform_1.zip"
-}
+[ 
+    "aws": {
```
hnicke: This was a simple example. The used name will be the name defined by the plugin itself. Later we could use a longer name, but only [a-z0-9-_] should be allowed (so resources can be used to build file or directory names on windows and unix file systems)

docs/draft/rest_api.md
```diff
@@ -1,201 +1,293 @@
...
+        "href"="/platforms/aws"
+    },
+    "cloudformation": {
+        "href"="/platforms/cloudformation"
+    }
+]
 ```
-- *id*: The id of the transformation (unsigned Integer)
-- *result*: the URL path (on the HTTP server) to download the result archive. If the transformation is still in progress or has errored this value will be null
+NOTE: Later it is possible to add platform specific preferences here
```
hnicke: why not. though i'd add it later, at the point when we see the arising need of preferences

docs/draft/rest_api.md
```diff
@@ -1,201 +1,293 @@
...
+```
+*returns:*  
 ```json
-{
-  "message": "Aborting transformation did not work.",
-  "error_code": 121341,
-  "log": "log"
-}
+[
+    "Database Password": {
```
hnicke: You mean without the space character? I thought about allowing spaces in the key values, in order to display them nicely in the GUI! So in the gui client, no string transformation is needed.

docs/draft/rest_api.md
hnicke: Btw this alternative rest api draft is the old version, but refactored and enriched. So don't worry, I did not only take one look at the outdated version

docs/draft/rest_api.md
```diff
@@ -1,201 +1,324 @@
...
 
-This call returns the `'linecount'` last lines of the log. 
+##### GET /toscamodels/{appName}
+Get the TOSCA model which name matches {appName}.
 
-HTTP Response Code: 200
-  
+*returns*:
 ```json
 {
```
nfode: I would at a id-field our web app or cli might need ids to handle the transformations. Angular for example could take ids to optimize viewing lists etc.

docs/draft/rest_api.md
hnicke: fixed

docs/draft/rest_api.md
hnicke: fixed

docs/draft/rest_api.md
hnicke: appName would serve as id field since its an unique field. Having an id field would cause unnecessary bloat of data, IMO

hnicke: Regarding filtering and limiting data: We might need this later, but I would only add it when we see the need for it --  see the danger of bloating the api with overengineered functions

nfode: 👍

docs/draft/rest_api.md
```diff
@@ -1,201 +1,293 @@
...
- 
-*Parameters*:
-- *id*: The id of the transformation for which the logs should be retrieved
-- *linecount*: (Optional, 50 by default) The Maximum linecount to return. if this parameter is ``all`` all lines of the log will be sent
+*ERRORS*:  
+422 - "name" value already in use by other toscamodel
 
-**on success returns:**
+##### DELETE /toscamodels
+Delete all TOSCA models
```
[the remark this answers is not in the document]
nfode: yes

mmunozba:
The REST API now follows [Hypertext Application Language (HAL)](http://stateless.co/hal_specification.html).
This mainly means changes to links:
- Strict separation of links and properties
- Navigation through the API with ```"_links"```
- Main source through "self" link

There are still a few open issues though:
- **Embedding** (Should we use it?)
  - HAL allows embedding of resources (see [HAL Spec](https://tools.ietf.org/html/draft-kelly-json-hal-08) Page 8)
  - Resources can be sent with their embedded resources (e.g. transformation with artifacts, logs and properties requires 1 request instead of 4)
  - Can improve performance
  - Might make the implementation unnecessarily complex?
- **Logs**
  - The current specification of the logs aspect of transformation seems confusing to me
  - Should probably be changed according to HAL with next, find etc. (see [HAL Spec](HAL spec](https://tools.ietf.org/html/draft-kelly-json-hal-08) Page 6). 
This might also be needed for toscamodels if we expect a lot of models like @nfode suggested in the last comment on this PR.

Besides that I also did some changes to the formatting and phrasing which should be looked over one more time.

hnicke:
@mmunozba In the meeting we've decided to rename 'toscamodels' to ..i think it was 'csar'. could you please check the protocol and update the api regarding this issue, if you got the time.

hnicke:
I opt for renaming Transformations status 'user-input' into 'input-required'.

hnicke:
Can we please drop the 'unsigned int' type of a property? int should be sufficient. java natively doesnt support unsigned ints.

nfode:
> I opt for renaming Transformations status 'user-input' into 'input-required'.

I think 'required-input' is more fluent

hnicke:
imho required-input is wrong -- should be requires-input or input-required

mmunozba:
I took a look at the protocol and it seems like toscamodels should be named /csars or /apps and the current /csar should be renamed to /archive. 
I'd prefer to use /apps over /csars since it will also contain the transformations and calling it csar seems confusing to me, so:
`/apps/hello_world/transformations/aws` as opposed to `/csars/hello_world/transformations/aws`

Regarding `user-input`:
I think `input-required` is good since it describes the current status of the transformation and it's easy to understand. It's definitely more consistent with the other status messages compared to `user-input`.

I'll be implementing these changes right away.

c-mueller:
Looks good to me. Some minor things probably have to be changed, when we start implementing the api. Especially regarding the upload of a Large CSAR file.

mmunozba:
I ended up adding the embedding and log changes I thought about. The API should be much closer to HAL now. As @c-mueller said, we probably need to adjust a lot during the implementation, but the current version should suffice. At least in my opinion.

c-mueller:
Has been merged!
However dont remove the branch yet. Just in case

PR #38, opened 2017-09-25
Datamodel and component diagram

hnicke:
I've worked on a data model which could serve as a common base for the upcoming sprint (nr. 4).


hnicke:
unintentionally setup this pull request into the wrong direction (master -> architecture instead of architecture -> master)

PR #39, opened 2017-09-25
[WIP] Model of core component

hnicke:
I've worked on a data model which could serve as a common base for the upcoming sprint (nr. 4).


core/src/main/java/org/opentosca/toscana/core/data/Application.java
```diff
@@ -0,0 +1,23 @@
...
+     * Key is the platform identifier.
+     */
+    private Map<String, Transformation> transformations;
+    private Csar csar;
+
+    public Application(Csar csar){
+        this.csar = csar;
+    }
+
+    public void transform(Platform targetPlatform){
```
c-mueller: If i understood you correctly in the last meetings, this method will invorke the Transformator class to start the transformation asychronously. But how do we get access to the Transformator instance?

core/src/main/java/org/opentosca/toscana/core/data/LogEntry.java
```diff
@@ -0,0 +1,16 @@
+package org.opentosca.toscana.core.data;
+
+public class LogEntry {
+
+    public final long TIMESTAMP;
+    public final String MESSAGE;
+    public final LogLevel LOG_LEVEL;
+
+    LogEntry(long timestamp, String message, LogLevel level){
+        assert message != null;
```
c-mueller: Don't do this!
If the condition is false, the whole application will crash because it throws a AssertionError which is not handled anywhere (at least usually)

I would replace it with a IllegalArgumentException

core/src/main/java/org/opentosca/toscana/core/data/LogEntry.java
```diff
@@ -0,0 +1,16 @@
+package org.opentosca.toscana.core.data;
+
+public class LogEntry {
+
+    public final long TIMESTAMP;
```
c-mueller: I think this should be lower case letters.

All Capital Letters are usualy used for static variables/constants.

core/src/main/java/org/opentosca/toscana/core/data/Platform.java
```diff
@@ -0,0 +1,49 @@
...
+    public final Set<Property> requiredProperties;
+
+    /**
+     * Creates a new platform.
+     * @param id short identifier of platform. must match regex [a-z_-]+
+     * @param name displayable name of platform. must not be an empty string
+     * @param requiredProperties the properties the platform requires.
+     */
+    public Platform(String id, String name, Set<Property> requiredProperties){
+        assert id != null && id.matches("[a-z_-]+");
```
c-mueller: As said in the LogEntry Class use Exceptions instead of assert.

core/src/main/java/org/opentosca/toscana/core/data/Application.java
hnicke: Almost! This method will create a Transformation and add it to the list of transformations this application holds. This maybe needs to be renamed to createTransformation() 
But the transform() method in Transformation.java will call the Transformator. So, Transformation needs to know Transformator. Concluding, Application has to know Transformator. In the end, it might be the DataAccess class which has to know the transformator in the beginning?

core/src/main/java/org/opentosca/toscana/core/data/LogEntry.java
hnicke: Your right. I was unsure about naming conventions regarding final but non-static fields.
WIll change this

core/src/main/java/org/opentosca/toscana/core/data/LogEntry.java
hnicke: Your right, will replace asserts with runtime exceptions

core/src/main/java/org/opentosca/toscana/core/data/Platform.java
hnicke: fixed

core/src/main/java/org/opentosca/toscana/core/data/Application.java
hnicke: opened #40

core/src/main/java/org/opentosca/toscana/core/csar/Csar.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.Map;
+
+public interface Csar {
+
+    /**
+     * Returns a map of all transformation objects of this csar.
+     * This includes scheduled, ongoing and finished transformations.
+     * Key of each map entry is the platform identifier of its particular transformation.
+     */
+    public Map<String,TransformationImpl> getTransformations();
```
c-mueller: TransformationImpl or just Transformation?

i think its the second

core/src/main/java/org/opentosca/toscana/core/csar/Csar.java
```diff
@@ -0,0 +1,20 @@
...
+
+    /**
+     * Returns the identifier of the CSAR (which serves as identifier)
+     */
+    public String getIdentifier();
```
c-mueller: We also need a way to read the archive in this interface i would say something like:
```Java
public InputStream getCsarInputStream() throws IOException;
```

core/src/main/java/org/opentosca/toscana/core/csar/CsarService.java
```diff
@@ -0,0 +1,27 @@
...
+
+public interface CsarService {
+
+
+    /**
+     * Creates a new Csar instance
+     * @param csarStream the actual cloud service archive as InputStream
+     * @return the newly created Csar instance
+     */
+    Csar uploadCsar(InputStream csarStream);
```
c-mueller: The name of the archive has to be specified here. Reading it from the archive file is very diffcult if not impossible. 

The server will not know the original file name before uploading!

core/src/main/java/org/opentosca/toscana/core/transformation/Property.java
```diff
@@ -0,0 +1,54 @@
+package org.opentosca.toscana.core.transformation;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+public class Property<T> {
```
c-mueller: i would not use a generic type for Properties! This makes it impossible to differentiate passwords from regular strings.

I think the following solution could be good.
Instead of making Properties generic i would just store the values in Strings. Typing them is no good idea especially if we want to put them in a Properties file, which basically is a String to String Map.
The type would be stored in a PropertyType Enum containing the following types:
* ``TEXT`` - Just a regular string
* ``IDENTIFIER`` -  a Short string only containing lower case letters, numbers (not at the beginning) and dashes
* ``PASSWORD`` - A String that is "confidential" and should not be shown in cleartext on screen. This mostly is a frontend indicator to tell the CLI or GUI to hide the input with `*`
* ``URL`` -  A url that will point to something
* ``NUMBER`` - A Integer Number (Internal Datatype will probably be Long)
* ``FLOAT`` - Floating Point Number (Internal Datatype: Double)

The propery class will contain the following attributes (when following my approach):
```Java
//This gets set to true if the property is required for transformation
//However replacing it with a enum to differentiate between a Required Property to run the Transformation or a Deployment Property
private boolean required; 
//The type of the Property
private PropertyType type;
//The identifying key of the property
private String key;
```
This approach also seperates the schema of the properties and their values Storing the values can be aly easy as putting them into a seperate Map<Property, String> this could also be wrpapped somehow tough.

core/src/main/java/org/opentosca/toscana/core/csar/Csar.java
hnicke: of course, fixed

core/src/main/java/org/opentosca/toscana/core/csar/Csar.java
hnicke: I thought about:
- forwarding the inputstream to CsarDao
- CsarDao persists archive to disk
- CsarDao unzips archive (maybe also deletes original archive to save disc space). [unsure about this step: today i try to integrate opentosca's parser, then i might know more]
- CsarDao gives File references to Csar. We don't want to work with inputstreams but use the File API.

Keep in mind this is all [WIP]

core/src/main/java/org/opentosca/toscana/core/csar/CsarService.java
hnicke: Alright. So now it's the clients responsibility to declare a name when creating a csar

c-mueller:
This pull request has moved to #52 to make integration with the rest of the core more easier

PR #41, opened 2017-09-28
Transformation workflow

hnicke:
Please review the new transformation workflow.
It it a result of the meeting of client Uwe, supervisor Michael, Nick and me.
I've further refined the resulting flow.


docs/dev/architecture/workflow.md
```diff
@@ -0,0 +1,31 @@
...
+- The plugins will always produce three kinds of output when transforming: First, the target model artifacts itself (e.g.: Cloud Formation Template and Deployment Artifacts).  
+Then, the `instance.properties` text file. It contains keys of required properties at deploy time (note: the values are unset).
+Last, they produce a deployment script `deploy.sh` which containts deployment and orchestration logic. When executing this script, it first sources the `instance.properties` text file, then deploys and orchestrates the app.
+    - This approach enables the user to transform any supported csar, then download the resulting artifacts and deploy at any given time with lightweight dependencies (transformator not required, only dependencies are shell & basic unix tools & targetPlatform-specific tools).
+    - However, the user still has the possibility to declare his application-specific properties and deploy the app via the GUI / CLI. Note that internally, the above mentioned `instance.properties` file is filled with values and the `deploy.sh` script is executed.
```
c-mueller: Shell scripts might not be the best solution to provide platform independence. 
Maybe we could use a scripting language like python to execute the scripts. Or the plugin will supply something like a "deployment tool" that inplements the task in java.

docs/dev/architecture/workflow.md
```diff
@@ -0,0 +1,31 @@
+# Transformation workflow
+
+This document describes the current transformation workflow.
+Older versions are described afterwards.
+
+## Current Flow
```
c-mueller: This method seems to be the most viable one, however the problem with the platform dependant Shell Scripts has to be adressed. Maybe we can just supply a Batch script and a Shell script to guarantee platform independance

hnicke: Please reconsider the usage of Cygwin on windows.
Do you consider it a deal breaker to request windows user to install Cygwin for deployment? I estimate the percentage of dev/ops running Windows but not Cygwin on their work machine approaches 0.

docs/dev/architecture/workflow.md
hnicke: What bothers me is that if we use scripts (e.g. bash or python) we have to handle missing dependencies. simplest solution would be to tell user what is missing (e.g.: "missing aws-cli in classpath, aborting") and let the user install dependencies on their own. 

If we had a deployment tool written in java for every plugin (client coined this an "executor") I imagine building the plugins could become easier.
However, most important goal then would be to keep this executor lightweight.

hnicke:
We need a good but fast decision on this topic. please review.
Without knowing the outcome if this discussion, it's hard to build the core-plugin interface.

docs/dev/architecture/workflow.md
nfode: I think we can expect from a user using our tool to install missing dependencies.

> If we had a deployment tool written in java for every plugin (client coined this an "executor") I imagine building the plugins could become easier.
However, most important goal then would be to keep this executor lightweight.

I thought this was our plan?

hnicke: >I think we can expect from a user using our tool to install missing dependencies.

you're using manjaro and you wouldn't want any script to somehow install further dependencies -- you want that to happen through pacman. on ubuntu its yaourt, on windows there's none -- so how would you handle this..?

> I thought this was our plan?

This was the clients vision after talking about the issue for 10 minutes. I tried to refine the idea even further. Please read the document again. My vision is to let the transform plugins create a `deploy.sh`, and have a single executor in the app which only has to run this script.

nfode: I meant with a proper doc about what dependencies our tool needs everyone can install the by themselves.

>This was the clients vision after talking about the issue for 10 minutes. I tried to refine the idea even further. Please read the document again. My vision is to let the transform plugins create a deploy.sh, and have a single executor in the app which only has to run this script.

The client also told us if there is enough time at the end that it might be possible that we implement the feature to save the instance of a deployed application. This might not happen, but how do we do that with you `deploy.sh`. Is this logic then implemented in the executor?
I think if we only produce a `deploy.sh` it is not that useful for other use cases.

hnicke: >The client also told us if there is enough time at the end that it might be possible that we implement the feature to save the instance of a deployed application.

I think this is a misunderstanding. With our current flow, that feature is obsolete. I think the client mentioned this when we discussed version 1 of the flow.

nfode: > I think this is a misunderstanding. With our current flow, that feature is obsolete. I think the client mentioned this when we discussed version 1 of the flow.

This might be true. One thing that is not clear with your idea, do we get only a deploy script and the properties file or also a model? So the transformer produces a `deploy.sh`, `instance.properties`, `transformedmodel.whatever`?

Another thing. If the user needs to perform multiple actions to run our `deploy.sh` why do we not give him a executor that reduces the dependencies? So he wont have to install python or something that understands bash?

hnicke: >One thing that is not clear with your idea, do we get only a deploy script and the properties file or also a model?

See attached note to node 'transform'. A deploy.sh, instance.properties and an undefined amount of target model artifacts are generated by the transform plugin.

> If the user needs to perform multiple actions to run our deploy.sh why do we not give him a executor that reduces the dependencies?

An executor which reduces dependencies implies it contains these dependencies. This is not considered lightweight.

nfode: > An executor which reduces dependencies implies it contains these dependencies. This is not considered lightweight.

Not using Bash or Python is a dependency less or am I wrong?

hnicke: I'm not sure what kind of executor you mean; but i guess you mean a java executor, which introduces java dependency (i can live with that). This java executor would want to use some target platform specific sdks, e.g. aws sdk for java; to be portable that executor would contain these libs. Such an executor would be quite big (not my understanding of lightweight)

hnicke: i think if the transformation result is not executable with generic tools, the whole idea of having a complete transformation result even without deployment can be given up.

nfode: > I'm not sure what kind of executor you mean; but i guess you mean a java executor, which introduces java dependency (i can live with that). This java executor would want to use some target platform specific sdks, e.g. aws sdk for java; to be portable that executor would contain these libs. Such an executor would be quite big (not my understanding of lightweight)

Bash or Python might also need some sdks or tools pre installed. And how big is the transformer if his job is to create a executable script?

hnicke: Everyone can read a bash script and modify it if necessary.
Think about the  equivalent of `deploy.sh`. In which format would it be inm if the executor was a java executor? I guess it would be a custom format.

nfode: If we have a java executor you don't have a custom format you just feed him the transformed model and `instance.properties` file

hnicke: Unfortunately, you need means to communicate the non-transformable aspects of the CSAR to the executor -- he wants to realize these aspects via orchestration.

nfode: > Unfortunately, you need means to communicate the non-transformable aspects of the CSAR to the executor -- he wants to realize these aspects via orchestration

You are right.

But I don't think this discussion leads to anything so I am out of this.

c-mueller:
After thinking about this problem a bit i came across the following possible solution.

**Before that**: I want to clarify my opinion on platform independency. 
I dont think we can expect the user to have a shell (bash) for windows such as Cygwin, Git bash or the MinGW Shell installed.
Gathering dependencies on a shell script is also not too easy. Using this approach would also mean, that we have to implement the deployment operations twice once in Shell (or whatever) and once in java to perform a server side deployment. 

The fact that we have to implement these operations in java anyways brought the following idea into my mind.

Why don't we try to create a jar file that contains all the required operations to deploy the target artifact on the specific platform? This jar will be put into the archive of every created target artifact. 
The tasks of this jar would be the following

1. Read the `instance.properties` file 
2. Validate its contents
3. Perform the deployment

The problem is: How is something like this getting implemented?

Doing one deployer jar per platform plugin seems impossible (at least its way harder)

In order to realize this our application will have to be split into several maven modules:

Shown as a Component Diagram:
c2

The edges Describe the type of dependency:
* `Module1 -> Module2` means that Module1 depends on Module2
* `Module1 - Module2` represents a Cyclic Dependency (`Module1 -> Module2` and `Module2 -> Module1`

Modules:
- `core` - Containes the Server sided components (API, Persistance)
- `plugin-manager` - Provides Classes for managing the plugins, and interfaces for interaction (such as the csar interface)
- `aws`, `kubernetes`, `cloud-foundry` - The plugins, implementing transform and Deploy operations for a specific platform
- `deployer` -  The externalized deployment jar. (Specific code like main class...)

To prevent trouble with Cyclic Dependencies the Plugins and the Plugin Manager can be put into one module

When Compiling the "deployer" it takes all of the plugins code (including the transformation code) and puts them into a jar with the corresponding deployer module. This way will include a lot of unused code and the file will be pretty large because of many dependencies but it only needs a Java Runtime Environment to run the deployment. The user probably has installed a JRE, especially if he used the cli before

hnicke:
@c-mueller thanks for sharing your thoughts. I don't think your approach helps with our initial problem. Let me explain, one by one.

> Using this approach [shell scripts as transformation output] would also mean, that we have to implement the deployment operations twice once in Shell (or whatever) and once in java to perform a server side deployment.

If our transformation output is a shell script for deployment, we don't have to implement any deployment logic in java at all. Our java executor has only one capability: it can execute shell scripts. 

Let's step back. Why do we want to have a complete transformation artifact?
The underlying use case is something like that:
"As a user, I want to transform my TOSCA infrastructure code into an artifact of a concrete cloud technology in order to I. escape TOSCA technology lock-in or II. use technology-specific mechanism to advance the IoC."
This use case demands that the transformation artifact is in an open, modifiable format. A JAR is neither open nor modifiable, whereas a bash script fulfills given requirements.

We gain no real benefit if our transformation output format is a JAR. The only things we can do with that are:
- modify instance paramenters
- execute = deploy & orchestrate
What we can not accomplish is changing the orchestration part of the model, so concluding we can't change the model.

If this write-up didn't convince you, please consider a mumble meeting with me.

hnicke:
If platform independency is the only issue, i offer to implement code that converts the resulting bash script to batch.

c-mueller:
Because this has been approved by two people i will merge this PR into master

Issue #42, opened 2017-09-29
Basic Kubernetes plugin


nfode:
Check out how a simple TOSCA-model can be transformed into a Kubernetes artefact.

PR #53, opened 2017-10-11
Kubernetes basic plugin

mmunozba:
Based on #42.

Currently contains a basic plan for the automated workflow of the planned basic Kubernetes plugin.


docs/dev/kubernetes/kubernetes_basic_workflow.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Transformation
+- Data model of the simple app is provided by core
+- Application is already containerized through Docker
+  - **Note**: For the full version of the plugin an additional step is needed in order to containerize applications. This will probably be fairly complex, since the plugin must decide which parts of the application will be put in which container.
+- Plugin creates Kubernetes Deployment configuration (YAML) based on data model
+- **Output**: target artifacts (Deployment configuration, container)
+
+### Deployment
+- Plugin creates Kubernetes cluster
```
c-mueller: Really? Don't we just want to deploy on a existing K8s cluster?

docs/dev/kubernetes/kubernetes_basic_workflow.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Automated Workflow
+Describes the planned automated deployment of the simple application based on our [transformation workflow](https://github.com/StuPro-TOSCAna/TOSCAna/blob/transformation-flow/docs/dev/architecture/workflow.md).
+
+### Transformation
+- Data model of the simple app is provided by core
+- Application is already containerized through Docker
+  - **Note**: For the full version of the plugin an additional step is needed in order to containerize applications. This will probably be fairly complex, since the plugin must decide which parts of the application will be put in which container.
+- Plugin creates Kubernetes Deployment configuration (YAML) based on data model
+- **Output**: target artifacts (Deployment configuration, container)
```
c-mueller: The problem with kubernetes is, that it needs a image path on a registry in order to deploy the image on kubernetes. This point has to be adressed somewhere

docs/dev/kubernetes/kubernetes_basic_workflow.md
```diff
@@ -0,0 +1,29 @@
...
+- Packs applications in isolated containers instead of VMs (one application per container)
+- Deploys containers based on operating-system-level virtualization rather than hardware virtualization
+- Kubernetes mainly uses Docker for containers
+
+## Automated Workflow
+Describes the planned automated deployment of the simple application based on our [transformation workflow](https://github.com/StuPro-TOSCAna/TOSCAna/blob/transformation-flow/docs/dev/architecture/workflow.md).
+
+### Transformation
+- Data model of the simple app is provided by core
+- Application is already containerized through Docker
```
c-mueller: Does that mean it's already in a registry?

docs/dev/kubernetes/kubernetes_basic_workflow.md
mmunozba: I'm not sure if I understand this properly. 

We have a container image and a Kubernetes cluster. In order to deploy the image on this cluster, we must create a Deployment configuration (YAML) file which specifies the location of our container image and other deployment related information. 
So is this location the one which is the path to the container image on the registry? 

Also, this registry is hosted seperately from the Kubernetes cluster and our TOSCAna application, right?

docs/dev/kubernetes/kubernetes_basic_workflow.md
mmunozba: Yes, after thinking about it, making the plugin create a kubernetes clusters is probably a bad idea. It would require remote creation of the cluster (or else the cluster would be hosted where TOSCAna is running I believe?). You should definitely be able to deploy on existing clusters, this should be changed.

docs/dev/kubernetes/kubernetes_basic_workflow.md
mmunozba: From my understanding, the application is a TODO app inside of a Docker container, so I assume it wouldnt be hard to get the container image. I commented below regarding the registry issue.

docs/dev/kubernetes/kubernetes_basic_workflow.md
c-mueller: Because docker does not feature a format to transport images as files there are only two ways to "move" a docker image to a different machine. The first one is taking the "source" of the docker image and building it whereever we need it.
The other option is to push the docker Image onto a registy and the machine on wich you want to access the image just pulls from the registry. This is the way to go because kubernetes only supports pulling from registires.
For the transformation we have to differenatiate between two different cases:
**First**:
the csar describes a path in a registry. In this case we dont have to do anything except for linking the image in the Kubernetes deployment. (Ignoring authentication for now) because kubernetes will deploy it from this image path.
**Second**:
the csar only features the "source" of the Image. Here we have to build the image (using `docker build` on a appropriate host) and then push it to a registry that has to be supplied by the user. After pushing we got a "URL" that we can use in combination with kubernetes.

To your last question:
Usualy the registry is hosted completely seperately,

nfode:
Added:
- how to setup kubernetes (checkout docs/dev/kubernetes)
- basic kubernetes config file with a ReplicationController and Service to test our plugin

c-mueller:
I would not use a Replica Set / Replication Controller. A deployment and a service is probably the best solution.

nfode:
+ Added a draft for a simple k8s-transformation-plugin how we discussed yesterday.
+ Added a basic roadmap for the kubernetes plugin

nfode:
@c-mueller @hnicke This can also be merged. Tagging you because you are our current merge masters 😄

nfode:
Can someone please take a look at this?

PR #56, opened 2017-10-13
Add ADR for extending the TOSCA YAML model

koppor:
This PR adds the result of the discussion mainly @hnicke and me had about the TOSCA YAML data model.

<s>It also adds [MADR](https://github.com/adr/madr), which should be used to document architectural decisions. More information on architectural decision records is available at <http://adr.github.io/>.</s>

### DIFF
--- a/docs/adr/0001-extend-tosca-yaml-model.md
+++ b/docs/adr/0001-extend-tosca-yaml-model.md
@@ -1,0 +1,45 @@
+# Extend TOSCA YAML model to a full in-memory model
+
+As a programmer, I want to access the complete YAML model and not be forced to traverse through all files.
+
+For instance, `TNodeTemplate.java` offers `public QName getType() {`, but one wants `public TNodeType getType()`.
+
+## Considered Alternatives
+
+* Inherit from plain data model and add transparent resolution
+* New data model
+* Eclipse EMF
+
+## Decision Outcome
+
+* Chosen Alternative: Inherit from plain data model and add transparent resolution
+* Comes out best (see below)
+
+
+## Pros and Cons of the Alternatives <!-- optional -->
+
+### Inherit from plain data model and add transparent resolution
+
+A new model with all classes in a different package inheriting from all model classes.
+Each class gets new methods.
+For the type example:
+
+- `public TNodeType getTypeResolved()`
+- `public setType(TNodeType nodeType)`
+
+This model is offered by a new repository implementation, which resolves the objects on demand.
+
+* `+` Transparent for the user
+* `+` `org.eclipse.winery.model.tosca.yaml` reused
+* `-` EMF Resource Set handling is implemented manually
+
+### New data model
+
+* `+` Clean data model
+* `-` Much code from `org.eclipse.winery.model.tosca.yaml` duplicated
+
+### Eclipse EMF
+
+* `+` [Resource Sets](http://download.eclipse.org/modeling/emf/emf/javadoc/2.5.0/org/eclipse/emf/ecore/resource/ResourceSet.html) handling does this transparent solution
+* `-` Changing all the model to EMF causes huge overhead
+* `-` Getting EMF into non-OSGi projects is hard


docs/adr/0001-extend-tosca-yaml-model.md
```diff
@@ -0,0 +1,45 @@
+# Extend TOSCA YAML model to a full in-memory model
+
+As a programmer, I want to access the complete YAML model and not be forced to traverse through all files.
```
hnicke: This does not catch what we are trying to achieve.
Steps to be done are (in this exact order):
1. Merge TServiceTemplates
- every parsed yaml file produces its own TServiceTemplate
- output of this operation shall be one unified TServiceTemplate
2. Resolve Types
- for every TNodeTemplate and TRelationshipTemplate, inject the members of their corresponding Types into the Template object (this is meant within this review by the term 'resolve')
- again, output of this operation shall be a TServiceTemplate
3. Write AccessibilityWrapper
- Supply plugins with any needed logic to operate on the model, e.g.
  - easy traversal of model, e.g. supply NodeTemplateStacks
  - ... (im sure this will grow from time to time)

docs/adr/0001-extend-tosca-yaml-model.md
```diff
@@ -0,0 +1,45 @@
...
+### Inherit from plain data model and add transparent resolution
+
+A new model with all classes in a different package inheriting from all model classes.
+Each class gets new methods.
+For the type example:
+
+- `public TNodeType getTypeResolved()`
+- `public setType(TNodeType nodeType)`
+
+This model is offered by a new repository implementation, which resolves the objects on demand.
```
hnicke: I think this won't work (or I don't understand what you want to do) nor is it neccessary. 
Why do we need a repository? All yaml template files get parsed, merged into one TServiceTemplate, and then get resolved in-memory. Anything I miss?

docs/adr/0001-extend-tosca-yaml-model.md
```diff
@@ -0,0 +1,45 @@
...
+
+* Chosen Alternative: Inherit from plain data model and add transparent resolution
+* Comes out best (see below)
+
+
+## Pros and Cons of the Alternatives <!-- optional -->
+
+### Inherit from plain data model and add transparent resolution
+
+A new model with all classes in a different package inheriting from all model classes.
```
hnicke: I don't think we have to extend the model.
For resolving, we can traverse the merged (after step 1) TServiceTemplate and compare qnames.
And for the plugins, the qname/simple name of the node template's type is sufficient (after nodetypes are resolved)

docs/adr/0001-extend-tosca-yaml-model.md
koppor: The repository implementation can rely on files stored in a file system. Just a simple repository which parses ALL the files and returns the necessary objects.

docs/adr/0001-extend-tosca-yaml-model.md
koppor: Keep in mind that TServiceTEmplate of TOSCA YAML equals to TDefinitions in TOSCA XML.

For the programmer, there should be no difference, how the enriched objects are returned.

docs/adr/0001-extend-tosca-yaml-model.md
koppor: If that is enough, you do not need to touch the model at all. Just implement a lightweight repository, which parses all `.yml` files in a repository and offer a QName resolution.

With the QName resolution, the question is, where to cast to TNodeType / TRelationsipType.

`public Object resolve(QName qname)` or `public TNodeType resolveNodeType(QName qname)` or `public TNodeType resolve(NodeTypeId)`.

In the last case, the caller has to create an appropriate Id. E.g., new NodeTypeId(qname). The caller knows that he wants to resolve a NodeType's QName.

In winery, we do the last option. See: https://github.com/eclipse/winery/blob/1dfac23e558fdc7bcd2122a4d98c89e54d16a3f0/org.eclipse.winery.common/src/main/java/org/eclipse/winery/common/interfaces/IWineryRepositoryCommon.java#L84

`default TNodeType getElement(NodeTypeId id) {`

docs/adr/0001-extend-tosca-yaml-model.md
koppor: I always have a "Repository" in mind, which is the "AccessibilityWrapper" in your terms. The Repository can be implemented using your steps 1 and 2. It can also implemented differently. The crucial point is how to treat the QNames.

hnicke:
This is outdated. Further discussion in #171

PR #52, opened 2017-10-05
[WIP] Implementation of the Core Component

c-mueller:
In order to make the development of the core component more seamless @hnicke and I have decided to merge our branches into one to work on the core component.

This Pull Request combines #39, #51 and #58.

The following tasks have to be done on the REST-API: 
- [x] Implement `/status` Request
- [x] Implement `/platforms` Request
- [x] Implement `/platforms/{id}` Request
- [x] List CSARs
- [x] Implement CSAR Upload
- [x] Create a Transformation
- [x] List all Transformations for a platform
- [x] Provide Target Artifacts as static files
  - [x] Implement API Call
  - [x] Implement handling mechanism providing this artifact folder
- [x] Implement a Way to get Logs through the API
- [x] Implement CsarController and the Corresponding Requests
- [x] Rework User Property Input mechanism (see #55)
  - [x] Rework Data Model
  - [x] Document new Approach
  - [x] Create Unit tests for the properties
- [x] Update REST API Specification (if needed)
   - Has been Replaced by Javadoc Comments, the `samples.md` file and the swagger Docs provided by springfox.
- [x] Replace Dummies with true implementations (once they are done, See #39)
    - Has been done, event though the components are not fully functional yet. No dummies get used (except for testing) 
- [x] Document the implemented classes
- [x] Implement unit tests for all requests implemented

Other tasks, regarding transformation handling (taken from #58)
**Common**:
- [x] Implement Transformation class
- [x] Implement TransformationService class
- [x] Document implemented code
- [x] Create unit tests

**Plugin Handling**
- [x] Create basic plugin interface
  - [x] define interface
  - [x] document interface
- [x] Implement Class/Mechanism to manage plugins (will also implement the PlatformService)
- [x] Find a mechanism for plugin loading 
	- Possible mechanisms
		- Spring Dependency Injection (Probably the best approach, because its "dynamic" and does not require manual class path scanning, implementation is also quite simple) **CHOSEN**
		- Static definition of the plugins in a array
		- Manual annotation processing
- [x] Document implemented code
- [x] Create unit tests

**Scheduling Transformations**
- [x] Implement auto-scheduling mechanism
- [x] Document implemented code
- [x] Create unit tests

Once finished this PR will implement #45, #50

PR #58, opened 2017-10-18
[WIP] Core-Component Transformation handling

c-mueller:
This PR will implement the handling of transformations and platforms meaning:
- Automaically starting (Scheduling) Transformations
- Handling the Logging within a Transformation
- Plugin handling

To Realize this, the following tasks have to be completed

**Common**:
- [ ] Implement Transformation class
- [ ] Implement TransformationService class
- [ ] Document implemented code
- [ ] Create unit tests

**Plugin Handling**
- [ ] Create basic plugin interface
  - [x] define interface
  - [ ] document interface
- [ ] Implement Class/Mechanism to manage plugins (will also implement the PlatformService)
- [x] Find a mechanism for plugin loading 
	- Possible mechanisms
		- Spring Dependency Injection (Probably the best approach, because its "dynamic" and does not require manual class path scanning, implementation is also quite simple)
		- Static definition of the plugins in a array
		- Manual annotation processing
- [ ] Document implemented code
- [ ] Create unit tests

**Scheduling Transformations**
- [ ] Implement auto-scheduling mechanism
- [ ] Document implemented code
- [ ] Create unit tests

More tasks will follow as the implementation proceeds


c-mueller:
Merged into #52

Issue #61, opened 2017-10-19
Handle Transformation target artifacts


c-mueller:
TBD

PR #80, opened 2017-10-25
Implemented Artifact Handling

c-mueller:
The PR Implements Issue #61. At least mostly it currently does not store the resulting URL because the Transformation does not have a way to set the TargetArtifact that does not cause a NullPointerException. Maybe @hnicke can have a look at this problem.

Tests have also been written using the TransformationServiceTest.

One Question is still open in this field. Do we want to compress the "artifacts" if a transformation fails and offer them for "debugging" purposes? My current implementation is handling it this way!

### DIFF
--- a/core/src/main/java/org/opentosca/toscana/core/api/ArtifactController.java
+++ b/core/src/main/java/org/opentosca/toscana/core/api/ArtifactController.java
@@ -1,7 +1,19 @@
 package org.opentosca.toscana.core.api;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.OutputStream;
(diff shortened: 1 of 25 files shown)


hnicke:
- fixed the nullpointer problem: TargetArtifact was never even created
- refactored the whole thing
  - renamed service
  - moved artifactdir from Preferences to Service
- solved some DI problems (circular dependencies of DAO's), hopefully for the last time. Ditched `@Lazy` annotation in favour of `@PostConstruct` approach
- redesigned error handling of ExecutionTask. Before, if transformation threw an error, artifact was not served. Anyways, i like your proposal to serve an target artifact even if transformation fails
- if transformation does not fail but does not generate output files, transformations state will now be set to ERROR. fixed tests.

PR #84, opened 2017-10-25
Refine Testing mechanisms

c-mueller:
This PR Updates the "CI Aware" Junit runners to allow the categorisation of tests.  The tests can now be Categorized in three categories:

The execution Mode gets selected by the environment variable `TEST_MODE` if this variable is not set, all tests will be executed

1. `TestModes.UNIT` - "Unit"-Tests, these tests will be executed by travis, in the default setup. For this mode TEST_MODE has to be set to `ci` (Cases get ignored in the value)
2. `TestModes.INTEGRATION` - Represents a middle level for "Integration" tests, because of the implementation style of our unit tests this might be a unecessary category and might be removed. TEST_MODE has to be set to `integration`
3. `TestModes.SYSTEM` - Represents System tests (Like the upload test). TEST_MODE has to be set to `integration`

If TEST_MODE gets set to one of the categories above, all tests from the other categroies will be ignored. To run all tests Don't set test mode or set the value to `all`.

For this categroisation a special JUnitRunner has to be used:
```
ModeAwareJUnitRunner - Extends the regular JUnit4 Runner
ModeAwareSpringRunner - Special Runner for spring based tests (Added to BaseSpringTest)
```

With these runners set you can set the Category of a test (Class or method) using this annotation:
`@TestMode(TestModes.UNIT)` - This test will only run in the Unit mode
Examples:

On a TestClass:
```Java
@RunWith(ModeAwareJUnitRunner.class)
@TestMode(TestModes.UNIT)
public class ZipUtilityTest {
```

On a Test Method (The runner still has to be one of the ones shown above):
```Java
@TestMode(TestModes.SYSTEM)
@Test(timeout = 30000)
public void testFileUpload() throws Exception {
```

Current Limitation: Parameterized tests cannot be Categorized yet! Also: these mechanisms only work within the core module!

I Also added a Jenkins Pipeline that executes the three types of tests one after the other. This is going to be usefull once the plugins have complex tests that might need the execution of a deployment platform like Kubernetes.


c-mueller:
Yes thats possible

PR #102, opened 2017-10-27
Refactored K8sPlugin test

nfode:
Forgot to refactor the `transformTest()` in the K8sPluginTest.


core/src/test/java/org/opentosca/toscana/plugins/k8s/K8sPluginTest.java
```diff
@@ -41,11 +42,10 @@
     @Autowired
     TestTransformationContext testTransformationContext;
```
hnicke: Limit the scope of fields: `private` would be appropriate.

hnicke:
Due to applied fix 1ef422d786296f2bb611de985ba7248ebb0c4791 the transformTest() fails.
Currently blocked by #103 and #104.

core/src/test/java/org/opentosca/toscana/plugins/kubernetes/KubernetesPluginTest.java
```diff
@@ -0,0 +1,96 @@
...
+    @Test(expected = InvalidDockerAppException.class)
+    public void dockerAppMissing() throws FileNotFoundException, InvalidCsarException, InvalidDockerAppException {
+        Csar csar = testCsars.getCsar(TestCsars.CSAR_YAML_INVALID_DOCKERAPP_MISSING);
+        tServiceTemplate = csarParser.parse(csar);
+        plugin.getDockerApp(tServiceTemplate);
+    }
+
+    @Test
+    public void transformTest() throws Exception {
+        File artifactRootPath = new File(preferences.getDataDir(), "test-csar/transformations/testplatform_one/");
```
hnicke: Your test should not hard code the artifact path. When the repo implementation changes, your test breaks.
As your using the TestTransformationContext, I suggest you use:
```
@Autowired
TransformationDao transformationDao;
transformationDao.getContentDir()
```

core/src/test/java/org/opentosca/toscana/plugins/kubernetes/KubernetesPluginTest.java
```diff
@@ -0,0 +1,96 @@
...
+        TransformationContext transformationContext = testTransformationContext.getContext(CSAR_YAML_VALID_DOCKER_SIMPLETASK, TestPlugins.PLATFORM1);
+        assertNotNull(transformationContext);
+        plugin.transform(transformationContext);
+
+        checkFilesInTransformationRoot(artifactRootPath);
+        checkDockerFilesInTransformationRoot(artifactRootPath);
+    }
+
+    private void checkDockerFilesInTransformationRoot(File artifactRootPath) {
```
hnicke: Look, this is a perfect example: Yesterday, the TransformationDao implementation changed.
Transformation Artifacts (not the resulting archive, but the single files) are now copied to the subdir /contents in the transformationRootDir. If you need the paths, please use transformationDao.getContentDir() or transformationDao.getRootDir() (here you will find the .zip)

core/src/test/java/org/opentosca/toscana/plugins/kubernetes/KubernetesPluginTest.java
```diff
@@ -0,0 +1,96 @@
...
+        List<String> result = new LinkedList<>();
+        result.clear();
+        for (String s : transformationDockerFilesPath.list()) {
+            result.add(s);
+        }
+        assertTrue(expectedDockerPathFiles.containsAll(result));
+    }
+
+    private void checkFilesInTransformationRoot(File artifactRootPath) {
+        List<String> expectedTransformationRoot = Lists.newArrayList("simple-task-app", "simple-task-app_resource.yaml", "Readme.md", "test-csar-testplatform_one.log");
```
hnicke: I would completey change the philosophy of the test: I suggest you mock the PluginFileAccess and verify that the methods got called. E.g., verify that `accessMopluginFileAccess.copy("simple-task-app")` got called.

core/src/test/java/org/opentosca/toscana/plugins/kubernetes/KubernetesPluginTest.java
nfode: So you suggest changing the TestTransformationContext class?

core/pom.xml
```diff
@@ -31,6 +31,12 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.mockito</groupId>
```
c-mueller: Is this necessary? Doesn't spring boot test ship with Mockito?

docs/dev/api/api_samples.md
```diff
@@ -79,13 +79,13 @@
 	"_embedded": {
 		"csar": [
 			{
-				"name": "k8s-cluster",
+				"name": kubernetes,
 				"_links": {
 					"self": {
-						"href": "http://localhost:8080/csars/k8s-cluster"
+						"href": kubernetes
```
c-mueller: Hmm this is not the proper value.

core/pom.xml
nfode: I have to retest but the version shipped with spring boot was to old to mock final classes.

docs/dev/api/api_samples.md
nfode: Ou something with refactoring went wrong.

core/pom.xml
nfode: Yep I was right, the newes version is needed to mock `final` classes. In my case the `TransformationContext`

Issue #103, opened 2017-10-27
Handle unsupported plattform for  Transformation


hnicke:
The core crashes when reading a transfromation from disk, if the transfromation's platform is not supported by any plugin.

PR #106, opened 2017-10-27
Fix create transformation with unsupported platform

hnicke:
Resolves: #103
- Refactor testsuite: PluginDummies now use Platforms defined in
TestPlugins class
- Transformations on disk which belong to unsupported platform will
not crash application anymore. Invalid transformation folder gets
deleted instead
- Trying to create a transformation with an unsupported platform
throws UnsupportedPlattformException


core/src/main/java/org/opentosca/toscana/core/plugin/AbstractPlugin.java
```diff
@@ -1,56 +1,48 @@
...
 
-    public AbstractPlugin() {
+    public AbstractPlugin(Platform platform) {
+        this.platform = platform;
+        if (platform == null) {
+            throw new IllegalStateException("initPlatform() must return the supported Platform of the plugin.");
+        }
         this.init();
-        logger.info("Initialized '{}' Plugin.", getName());
+        logger.info("Initialized '{}' Plugin.", platform.name);
```
c-mueller: Maybe we should hide this information. Would probably look better. if thats not done already. this attribute should at least be final to prevent modification

core/src/main/java/org/opentosca/toscana/core/plugin/AbstractPlugin.java
```diff
@@ -1,56 +1,48 @@
...
-     */
-    public abstract String getIdentifier();
-
-    /**
-     * @return The plugin specific properties like enpoints get provided by this mehtod
+     * @return the supported platform of the plugin
      */
-    public HashSet<Property> getPluginSpecificProperties() {
-        return new HashSet<>();
+    protected Platform initPlatform() {
```
c-mueller: This is not needed anymore when we require the insertion of the platform object into the constructor.

core/src/main/java/org/opentosca/toscana/core/transformation/TransformationDao.java
```diff
@@ -19,18 +20,19 @@
      * @param csar     the csar on which the new transformation will be based on
      * @param platform the target platform of the new transformation
      * @return the new transformation
+     * @throws UnsupportedPlatformException if given platform is not supported by any known plugin
      */
-    Transformation create(Csar csar, Platform platform);
+    Transformation create(Csar csar, Platform platform) throws UnsupportedPlatformException;
```
c-mueller: Does the frontend (rest api) need to handle this exception?

core/src/main/java/org/opentosca/toscana/core/transformation/UnsupportedPlatformException.java
```diff
@@ -0,0 +1,19 @@
+package org.opentosca.toscana.core.transformation;
+
+import org.opentosca.toscana.core.transformation.platform.Platform;
+
+/**
+ * Gets thrown when a platform is encountered which is not supported by any known plugin
+ */
+public class UnsupportedPlatformException extends Exception {
```
c-mueller: This is just a sidenote:
If  this has to be handled by the API: add `@ResponseStatus(value = <Error Code> = <Some message>)` here to tell spring what to respond when this exception gets thrown

core/src/main/java/org/opentosca/toscana/core/api/TransformationController.java
```diff
@@ -145,11 +146,17 @@ public TransformationController(CsarService csarService,
...
         if (p == null) {
             throw new PlatformNotFoundException();
         }
-        Transformation transformation = transformationService.createTransformation(csar, p);
+        Transformation transformation = null;
+        try {
+            transformation = transformationService.createTransformation(csar, p);
+        } catch (UnsupportedPlatformException e) {
+            log.error("Plattform {} not supported by any platform", e.platform, e);
+            throw new PlatformNotFoundException();
```
hnicke: @c-mueller here the exception is handled. Did I patch it the right way?

core/src/main/java/org/opentosca/toscana/core/transformation/TransformationDao.java
hnicke: yes - look at the patch in TransformationController

PR #107, opened 2017-10-27
Rework properties Request Mappings and property behaviour

### LINKED ISSUE #109: PropertyInstance should update TransformationState
In order to detect if all properties (the required ones) have been set the Property instance should change the transformation state to READY (Currently known as CREATED) to allow the launching of the transformation.

c-mueller:
This PR modifies the `/properties` response mappings for transformations in the way we discussed in the last meeting.

This means:
- `GET /properties` now also returns the current value of the property (null if it is not set). It can also be called when the transformation is not in the `INPUT_REQUIRED` state anymore.
- `PUT /properties` now returns a empty body if all sent properties have been set sucessfully
    - The original response only gets sent if at least one of the set values is invalid

The reason why this is "WIP":
- Maybe we should change the state to created once all required properties have been set. To inform the user once all required properties have been set

This PR Also resolves #109, #110 (was not much todo)

### DIFF
--- a/core/src/test/java/org/opentosca/toscana/core/api/TransformationControllerTest.java
+++ b/core/src/test/java/org/opentosca/toscana/core/api/TransformationControllerTest.java
@@ -12,10 +12,9 @@ import org.opentosca.toscana.core.dummy.DummyCsarService;
 import org.opentosca.toscana.core.dummy.DummyPlatformService;
 import org.opentosca.toscana.core.dummy.DummyTransformation;
 import org.opentosca.toscana.core.dummy.DummyTransformationService;
-import org.opentosca.toscana.core.transformation.TransformationService;
-import org.opentosca.toscana.core.transformation.TransformationState;
 import org.opentosca.toscana.core.transformation.platform.PlatformService;
 
+import org.json.JSONArray;
 import org.json.JSONObject;
 import org.junit.Before;
 import org.junit.Test;
@@ -24,9 +23,13 @@ import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.MvcResult;
 import org.springframework.test.web.servlet.setup.MockMvcBuilders;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
+import static org.opentosca.toscana.core.transformation.TransformationState.TRANSFORMING;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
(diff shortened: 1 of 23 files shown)


core/src/test/java/org/opentosca/toscana/core/api/TransformationControllerTest.java
```diff
@@ -69,6 +71,42 @@ public void setUp() throws Exception {
...
+        mvc.perform(
+            get("/csars/k8s-cluster/transformations/p-a/start")
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(VALID_PROPERTY_INPUT)
+        ).andDo(print())
+            .andExpect(status().is(200))
+            .andExpect(content().bytes(new byte[0]))
+            .andReturn();
+        assertEquals(TRANSFORMING,
+            csarService.getCsar("k8s-cluster").getTransformations().get("p-a").getState());
```
hnicke: Please introduce fields for `platformName` and `sarName`, hardcoded Strings are hard to maintain.

core/src/test/java/org/opentosca/toscana/core/api/TransformationControllerTest.java
```diff
@@ -69,6 +71,42 @@ public void setUp() throws Exception {
...
+        mvc.perform(
+            get("/csars/k8s-cluster/transformations/p-a/start")
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(VALID_PROPERTY_INPUT)
+        ).andDo(print())
+            .andExpect(status().is(400))
+            .andExpect(content().bytes(new byte[0]))
+            .andReturn();
+        assertNotEquals(TRANSFORMING,
+            csarService.getCsar("k8s-cluster").getTransformations().get("p-a").getState());
```
hnicke: The newly introdcued fields can then be used here, too :)

core/src/test/java/org/opentosca/toscana/core/api/TransformationControllerTest.java
```diff
@@ -118,25 +152,45 @@ public void getTransformationProperties() throws Exception {
...
-    public void getTransformationPropertiesInvalidState() throws Exception {
+    public void getTransformationPropertyValues() throws Exception {
         preInitNonCreationTests();
-        ((DummyTransformation) csarService.getCsar("k8s-cluster").getTransformations().get("p-a"))
-            .setState(TransformationState.READY);
-        mvc.perform(
+        //Set a Property value
+        csarService.getCsar("k8s-cluster")
+            .getTransformations().get("p-a")
+            .getProperties().setPropertyValue("secret_property", "geheim");
```
hnicke: Please introduce variables for the strings, as they are used below for verification again.

core/src/test/java/org/opentosca/toscana/core/api/TransformationControllerTest.java
```diff
@@ -423,7 +485,13 @@ public void transformationDeleteCsarNotFound() throws Exception {
             .andDo(print()).andExpect(status().isNotFound()).andReturn();
     }
 
+    @Test
+    public void transformationStartCsarNotFound() throws Exception {
+        mvc.perform(get("/csars/keinechtescsar/transformations/p-a/start"))
```
hnicke: Please stick to english. Even Strings like these carry a meaning.

core/src/test/java/org/opentosca/toscana/core/transformation/properties/PropertyInstanceTest.java
```diff
@@ -0,0 +1,78 @@
...
+        for (int i = 0; i < 10; i++) {
+            properties.add(new Property("p-" + i, PropertyType.INTEGER, "", i < 5));
+        }
+        this.transformation = new DummyTransformation(new Platform("test", "test", properties));
+        this.instance = transformation.getProperties();
+    }
+
+    @Test
+    public void checkStateNoPropsSet() throws Exception {
+        assertEquals(INPUT_REQUIRED, this.transformation.getState());
```
hnicke: unnecessary `this`

Issue #104, opened 2017-10-27
Disk cleanup before and after tests suite


hnicke:
Sometimes, the test data is not properly wiped from the disk (when tests fail?)
- implement proper cleanup

hnicke:
Reopen because not yet resolved

PR #108, opened 2017-10-28
Refactor test suite

hnicke:
Resolves: #104
- Make Integration-scoped tests inherit from BaseSpringTest
- Make JUnit-scoped tests inherit from BaseJUnitTest
- Remove "dummy_plugin" `@Profile` annotation which was obsolete
- Add documentation for disk cleanup in base tests
- Fix disk cleanup in Base Test classes
- Refactor tests to use tmpdir from Base Test classes
- Change tmp root dir location for testsuite to folder in project dir


core/src/test/java/org/opentosca/toscana/core/BaseSpringTest.java
```diff
@@ -1,45 +1,64 @@
...
- * disk
+ * Configuration After every test method, refreshes the context. 
+ *
+ * If disk access is needed, use {@field #tmpdir}
  */
 @RunWith(CategoryAwareSpringRunner.class)
 @ContextConfiguration(classes = {TestCoreConfiguration.class})
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
-@ActiveProfiles("dummy_plugins")
 @Component
```
c-mueller: I honestly dont think we need this here, maybe run it without the annotation and see if its works

hnicke: you're right, removing `@Component`

core/src/test/java/org/opentosca/toscana/core/BaseJUnitTest.java
```diff
@@ -1,10 +1,34 @@
...
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.rules.TemporaryFolder;
 
+/**
+ * If files need to get written on disk during tests, use {@link #temporaryFolder}
+ */
 public abstract class BaseJUnitTest extends BaseTest {
```
c-mueller: Forgot this in the last review. We should add  `@RunWith(CategoryAwareJUnitRunner.class)`. Does not make any sense to not make the regular unit tests not category aware!

hnicke: fixed

PR #122, opened 2017-10-30
Use Spring Boot Actuator for metrics and move api in seperate directory

c-mueller:
Once implemented this PR will Resolve #111 and #74

What is done in this Pull Request?
- The original `/status` Mapping has been replaced by two Spring Boot Actuator mappings. Those beeing:
   - `/api/status/health` - Returning the current health status of the transformer and the filesystem. (Will add a sample response below). Reporting of the transformer status is done using the `TransformerHealthIndicator` this (kind of) replaces the old Status service and the FileSystem classes that have been removed.
     - This new implementation now shows the running/errored transformations instead of just saying that something is running or has errored. The loaded plugins also get listed.
   - `/api/status/metrics` - This is a mapping that i just took as it is provided by spring, it shows a lot of information about the system on which the transformer is running. More information can be found [here](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html)
- All of the Implemented mappings have been moved into the sub "directory" `/api`. If you want to get a list of all csars now you have to call `GET /api/csars` instead of `GET /csars`

Sample Responses (Comments will follow in the swagger documentation):
**Health Mapping**:
```Json
{
    "status": "UP",
    "transformer": {
        "status": "UP",
        "installed_plugins": [
            "kubernetes",
            "cloud-formation",
            "cloud-foundry"
        ],
        "running_transformations": [
            {
                "csar":"test",
                "platform":"kubernetes"
            }
        ],
        "errored_transformations": [
            {
                "csar":"test",
                "platform":"cloud-foundry"
            }
        ]
    },
    "diskSpace": {
        "status": "UP",
        "total": 113264439296,
        "free": 26136109056,
        "threshold": 10485760
    }
}
```

**Metrics Mapping**:
```Json
{
	"mem": 625667,
	"mem.free": 441807,
	"processors": 16,
	"instance.uptime": 161317,
	"uptime": 163915,
	"systemload.average": 0.39,
	"heap.committed": 571904,
	"heap.init": 258048,
	"heap.used": 130096,
	"heap": 3650048,
	"nonheap.committed": 55232,
	"nonheap.init": 2496,
	"nonheap.used": 53764,
	"nonheap": 0,
	"threads.peak": 40,
	"threads.daemon": 27,
	"threads.totalStarted": 74,
	"threads": 29,
	"classes": 7487,
	"classes.loaded": 7489,
	"classes.unloaded": 2,
	"gc.ps_scavenge.count": 7,
[...]

PR #123, opened 2017-10-30
Create a centralized Retrofit Wrapper

c-mueller:
I have already thought about this a while ago, Centralizing the Retrofit based API bindings allows updating of the api without affecting the CLI in any way at least if the required parameters and method names don't change.

Currently this just defines the Data Structure delivered in the rest api and a Interface (used by retrofit) to access the api.

Keep in mind that this one is closely bound to pr #122 which should be merged into master before this one.


hnicke:
Please fix (fixable) codacy issues before merging.

PR #126, opened 2017-10-30
Refactor/optional return

hnicke:
After screening how the api controllers have to handle calls to the service layer, I've wanted to change the service layers to always return the value wrapped in Optionals (java 8 feature) in case this value might be null. After the change, the api layer looks way cleaner.
I believe the usage of Optional will help us build more stable software.

Change:
- `csarService.getCsar(String): Csar` --> `csarService.getCsar(String): Optional<Csar>`
- `platformService.FindPlatformById(String): Platform` --> `platformService.FindPlatformById(String): Optional<Platform>`
- `csar.getTemplate()`: TServiceTemplate` --> `csar.getTemplate(): Optional<TServiceTemplate>`

Add:
- `csar.getTranformation(Platform): Optional<Transformation>`

### DIFF
--- a/core/src/main/java/org/opentosca/toscana/core/transformation/execution/ExecutionTask.java
+++ b/core/src/main/java/org/opentosca/toscana/core/transformation/execution/ExecutionTask.java
@@ -11,6 +11,7 @@ import org.opentosca.toscana.core.transformation.TransformationState;
 import org.opentosca.toscana.core.transformation.artifacts.ArtifactService;
 import org.opentosca.toscana.core.transformation.artifacts.TargetArtifact;
 
+import org.eclipse.winery.model.tosca.yaml.TServiceTemplate;
 import org.slf4j.Logger;
 
 public class ExecutionTask implements Runnable {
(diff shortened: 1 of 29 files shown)


core/src/main/java/org/opentosca/toscana/core/transformation/execution/ExecutionTask.java
```diff
@@ -72,7 +73,10 @@ private void serveArtifact() {
 
     private void transform() {
         try {
-            plugin.transform(new TransformationContext(transformation, csarContentDir, transformationRootDir));
+            TServiceTemplate template = transformation.getCsar().getTemplate().orElseThrow(() ->
+                new IllegalStateException("TServiceTemplate is null, will not start transformation."));
```
c-mueller: Maybe reword this to "TServiceTemplate of csar {} is null, could not start transformation"

hnicke: fixed

PR #141, opened 2017-11-03
Add Dockerfile for TOSCAna

c-mueller:
This dockerfile is meant for automatic building using dockerhub, It pulls the master branch and builds it within the container. Tests wont be executed within the build process.


c-mueller:
I Decided to directly rebase this. because the contents of the dockerfile are already known (See Dockerfiles Repo)

PR #142, opened 2017-11-03
Enhance Jenkinsfile with deploy stage

hnicke:
- clean will now always be performed (not as stage)
- when build and tests pass, the toscana server gets deployed.
  - endpoint is https://toscana.ddnss.org/api
  - a running instance gets killed before
  - each time the data dir will get a reset

As soon as this got merged into master, I will configure jenkins to build and deploy whenever there's a new commit on master.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,6 @@
 pipeline {
     agent any
+
     stages {
         stage('Build') {
             steps {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,51 +1,58 @@
...
+                            archiveArtifacts allowEmptyArchive: true, artifacts: '**/target/site/*.zip' 
+                        } 
+                    } 
                 }
             }
-        }
-        stage('Cleanup') {
+        } 
+        stage('Deploy') {
             steps {
```
c-mueller: Maybe add a When conditon here to only execute this if the Worker only has a specific label.

This will prevent the failiure of the build on systems that do not want do deploy automatically

Jenkinsfile
```diff
@@ -1,51 +1,58 @@
...
+                    environment { 
+                        TEST_MODE = 'fast' 
+                    } 
+                    steps { 
+                        sh 'export DATADIR=$(pwd)/toscana-data && rm -rf $(pwd)/toscana-data && mkdir $(pwd)/toscana-data && mvn test -B' 
+                    } 
+                    post { 
+                        always { 
+                            junit(testResults: '**/target/surefire-reports/*.xml', allowEmptyResults: true) 
+                            sh '[ -d core/target/site ] && cd core/target/site && zip -r coverage-core-fast.zip jacoco && cd -; exit 0'
```
c-mueller: Whatever this does. Here is the point where you see my bash skills :P

hnicke: take a look at `man test`. It's less black magic than you think.

Jenkinsfile
hnicke: fixed

PR #145, opened 2017-11-04
Fix Jenkinsfile

c-mueller:
The Deploy stage now only gets executed if a environment variable `TOSCANA_DEPLOY_ON_BUILD` is set to `true`. Otherwise this stage should be skipped. This change is necessary to let the build pass even if you do not want to deploy. The current approach with the label is okay, but jenkins will not finish the build if there is no Node with the label "deploy". Because it waits for a worker to appear with this label.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -44,7 +44,9 @@ pipeline {
             }
         } 
         stage('Deploy') {
-            agent { label 'deploy' }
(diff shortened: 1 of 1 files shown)


c-mueller:
No problem, did not know it either, was just wondered why the build is not finishing

PR #146, opened 2017-11-04
Clean up logs and refactor DaoTests

hnicke:
- Improve logging
  - suppress springfox logs
  - Log http requests in a nice way
- Refactor Dao Tests: They are now real unit tests (without spring)


core/src/test/java/org/opentosca/toscana/core/transformation/TransformationFilesystemDaoTest.java
```diff
@@ -24,41 +29,44 @@
...
 import static org.opentosca.toscana.core.testdata.TestPlugins.PLATFORM1;
 import static org.opentosca.toscana.core.testdata.TestPlugins.PLATFORM2;
-import static org.opentosca.toscana.core.testdata.TestPlugins.PLATFORM4;
+import static org.opentosca.toscana.core.testdata.TestPlugins.PLATFORM3;
 import static org.opentosca.toscana.core.testdata.TestPlugins.PLATFORM_NOT_SUPPORTED;
 
 public class TransformationFilesystemDaoTest extends BaseSpringTest {
 
     @Autowired
     private TestCsars testCsars;
```
c-mueller: Is this still used? Did not find any reference to this in code.

When removing this we could probably make this a Unit test too

hnicke:
fixed

PR #153, opened 2017-11-06
Target artifact conventions

nfode:
# Pull request for the target artifact conventions
This is at the moment nothing more than a collection of some ideas.
I am happy about input from the rest of the team.


c-mueller:
Just a little note:
I would suggest we stay with one spelling even though artifact and artefact are both correct (american and british english).

In the core we call the resulting file a `TargetArtifact` therefore i would suggest renaming `artefact` to `artifact`

nfode:
Update the draft with the things we settled on in our meeting. I am going to close and rebase this PR so I can create Issues for ther points that are still left to discuss.

PR #158, opened 2017-11-07
Add Plugin Planning document to master branch

c-mueller:


nfode:
Had a talk with @hnicke about this. We might start to just implement NodeTypes first. That means we have CSARs with a single node to start with. That means we might have to change this document, also in regard to the time we have left until the sprint review.

PR #159, opened 2017-11-07
Draft for a TransformationLifecycle

nfode:
This draft tries to solve the general plugin architecture problem by extracting as much logic as possible from the specific plugins itself to a more abstract level.


docs/draft/transformation-lifecycle/transformation-lifecycle.md
```diff
@@ -0,0 +1,25 @@
...
+With the lifecycle in mind we can build the TransformationLifecycleInterface containing a method for every build phase.
+The interface is then implemented by a abstract BasePlugin.
+Every specific plugin has to extend the BasePlugin, so the specific plugin is forced to provide a method for every transformation phase.
+
+**Transformation phases:**
+- `validate` - in this phase the plugin provides a list of supportet TOSCA types, then the graph that is enqued for transformation will be checked if he contains types the plugin can not handle.
+If the result is positive the transformation continues else it will abort.
+- `analyze` - in this phase the plugin analyzes the graph to find out how to handle it.
+For example the Kubernetes plugin in this case analyzes if there is for example a NodeTemplate that requires windows.
+- `processes` - in this step the plugin processes the graph to transform it in the following step. For example the kubernetes plugin needs to split the graph into container and pods.
```
c-mueller: Would rename this to `process`

docs/draft/transformation-lifecycle/transformation-lifecycle.md
```diff
@@ -0,0 +1,25 @@
...
+- `transformation` - this is where the real transformation is happening.
```
c-mueller: would rename this to `transform`

c-mueller:
The following is just "writing down what i was thinking"

The following class diagram just represents a possible variant of the interfaces i thought of. I was just thinking about how we could "implement" these processes
image
TLI is meant short for TransformationLifecycleInterface

I understood the document like this:
1. the `TransformationLifecycleInterface` (short TLI) gets implemented by every Plugin (possibly by inheriting a abstract version of it), just like the `TOSCAnaPlugin` (don't remember the current name of it in the code)
2. The TIA interface gets a instance for every running transformation. (this is probably the best way to implement this)

These "assumptions" result in the following ideas:
- The Plugin Interface (`TOSCAnaPlugin`, or whatever its called currently) represents a Generic Factory to instantiate a plugin specific version of the TIA interface (class, when implemented)
  - A alternative to this could be the returning of the class object (`<Classname>.class`) and instantiating it using the `newInstance()` method within the core, this method is not that great, because it requires the class to have a constructor with no parameters. (I know it is possible to invoke ones with parameters but this gets very ugly)
- The TIA Instance stores the data that gets gathered in the previous phases and possibly also the `TransformationContext` (this way the context does not have to be in every method header)
- The role of the `ExecutionTask` gets way more important because it will now toggle the instantiation of the TIA instance and also execute the tasks one by one.


The following Sequence Diagram just visualizes the process of running the transformation from a `core` perspective. I have simplified parts that were irrelevant for showing the possible execution sequence (`REST Call -> Transfomation Launch through service -> Execution`)

image

_A little sidenote:_ 
I was using [PlantUML](http://plantuml.com/) to create these diagrams the source files can be found [here](https://gist.github.com/c-mueller/0ed14c9df59551e90597657815431185)

docs/draft/transformation-lifecycle/transformation-lifecycle.md
```diff
@@ -0,0 +1,25 @@
...
+> **Disclaimer:** This is just a basic idea mainly by hnicke and developed by nfode. This is in an early stage and needs some discussion.
+
+The basic idea of the TransformationLifecycle is to create a interface that brings all the plugins on a same level.
+The gist of the TransformationLifecycle is to provide the same lifecycle for every transformation, the idea came from the Maven lifecycle. As in the Maven lifecycle we have different phases, in our case so called transformation phases.
+With the lifecycle in mind we can build the TransformationLifecycleInterface containing a method for every build phase.
+The interface is then implemented by a abstract BasePlugin.
+Every specific plugin has to extend the BasePlugin, so the specific plugin is forced to provide a method for every transformation phase.
+
+**Transformation phases:**
+- `validate` - in this phase the plugin provides a list of supportet TOSCA types, then the graph that is enqued for transformation will be checked if he contains types the plugin can not handle.
```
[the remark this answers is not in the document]
hnicke: Typo: did you mean `enqueued` instead of `enqued`?

docs/draft/transformation-lifecycle/transformation-lifecycle.md
```diff
@@ -0,0 +1,25 @@
...
+If the result is positive the transformation continues else it will abort.
+- `analyze` - in this phase the plugin analyzes the graph to find out how to handle it.
+For example the Kubernetes plugin in this case analyzes if there is for example a NodeTemplate that requires windows.
+- `processes` - in this step the plugin processes the graph to transform it in the following step. For example the kubernetes plugin needs to split the graph into container and pods.
+- `transformation` - this is where the real transformation is happening.
+- `package` - in this step the plugin copies the result of the transformation into the given folder.
```
hnicke: I don't think we need this phase. The plugins shall write data directly to the right spot with the help of PluginFileAccess in the `transform` phase.

docs/draft/transformation-lifecycle/transformation-lifecycle.md
hnicke: Please clarifiy that the validation itself is implemented in the baseplugin.

hnicke:
@c-mueller please don't comment on PR's in this way - there is no way for me to review your comment. I think this is approach is overcomplicating things.
- why do we need the factory pattern here?
- Maybe ExecutionTask should not know about lifecycle phases, but the BasePlugin itself should call its phases one by one when `transform()` is being called.

docs/draft/transformation-lifecycle/transformation-lifecycle.md
nfode: Yeah were not sure if I should add it but after some rethinking I have the same oppinion.

docs/draft/transformation-lifecycle/transformation-lifecycle.md
```diff
@@ -0,0 +1,23 @@
+# [WIP] Transformation Lifecycle draft
+
+The basic idea of the TransformationLifecycle is to create a interface that brings all the plugins on a same level.
+The gist of the TransformationLifecycle is to provide the same lifecycle for every transformation, the idea came from the Maven lifecycle. As in the Maven lifecycle we have different phases, in our case so called transformation phases.
+With the lifecycle in mind we can build the TransformationLifecycleInterface containing a method for every build phase.
+The interface is then implemented by a abstract BasePlugin.
+Every specific plugin has to extend the BasePlugin, so the specific plugin is forced to provide a method for every transformation phase.
+
+**Transformation phases:**
+- `validate` - in this phase the plugin provides a list of supported TOSCA types, then the graph that is enqued for transformation will be checked if it contains types the plugin can not handle.
```
JensMueller2709: just a question: 
the plugins support different node-types we defined [here](https://github.com/StuPro-TOSCAna/TOSCAna/blob/master/docs/draft/plugins/supported_types.md). But the node-types are the same for every plugin, the plugin only has to implement the special transformation of this type. Does it make sense to define every node-type in the BasePlugin? So that every plugin has access to the same node-types and not every plugin has to define the nodetypes by themselfes.

docs/draft/transformation-lifecycle/transformation-lifecycle.md
```diff
@@ -0,0 +1,23 @@
...
+If the result is positive the transformation continues else it will abort.
+- `analyze` - in this phase the plugin analyzes the graph to find out how to handle it.
+For example the Kubernetes plugin in this case analyzes if there is for example a NodeTemplate that requires windows.
+- `prepare` - in this step the plugin processes the graph to transform it in the following step. For example the kubernetes plugin needs to split the graph into container and pods.
+- `transform` - this is where the real transformation is happening.
+- `clean` - in this phase the plugin cleans up remaining leftovers from the previos steps.
```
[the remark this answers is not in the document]
hnicke: He thinks of files that are needed during transformation, but are not part of the resulting target artifact.

docs/draft/transformation-lifecycle/transformation-lifecycle.md
hnicke: This is the approach taken by #169 , check it out.

docs/draft/transformation-lifecycle/transformation-lifecycle.md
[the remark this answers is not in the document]
JensMueller2709: ok thanks, maybe it is useful to make it clear in the document.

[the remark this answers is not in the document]
nfode: Actually I wanted to implement my own garbage collection :smile:

nfode:
I think we can merge this PR because it is just a draft and a more detailed architecture can be discussed in #170.

docs/adr/0002-general-plugin-architecture.md
```diff
@@ -0,0 +1,31 @@
...
+
+
+## Pros and Cons of the Alternatives
+
+### Design a general plugin architecture
+
+- `+` every plugin implements the same interface
+- `+` similarities can be extracted to a more abstract level
+- `+` reduction of information the plugins need, for example file acces could be handled in the abstract layer
+
```
mklopp: `-` there may be things that are not required by every plugin

PR #161, opened 2017-11-08
Renamed Core to Server

### LINKED ISSUE #124: Change folder structure
The folder structure needs to be updated because the core package contains the real core but also the REST API. To ease communication an provide a better structure we decided to rename the core folder in the repository root to server and move the api package one level up.

`core` -> `server`
`/server/src/main/java/org/opentosca/toscana/core/api/` -> `/server/src/main/java/org/opentosca/toscana/api/`

c-mueller:
This PR only exists to ensure the renaming does not cause Travis to fail

It resolves #124


nfode:
> This PR only exists to ensure the renaming does not cause Travis to fail
It failed 😅

c-mueller:
Yes this occured when moving the api package out of the core package. i have no idea why this happened, did revert this part for now. I will open a new issue to resolve this part of the original issue

nfode:
Did it run locally?

c-mueller:
No thats not the pom thats a problem with annotation scanning.

PR #169, opened 2017-11-09
Design of EffectiveModel

### LINKED ISSUE #168: Design common instance model
We need a common model that represents the topology of a CSAR. This model shall be used by the plugins to execute the transformation.

hnicke:
We need a common model that represents the topology of a CSAR. This model shall be used by the plugins to execute the transformation.

Will resolve #168 upon completion.

- [x] maybe also add all non-normative Types (TOSCAV1.1, ch. 9).


docs/dev/effective-model/effective-model.md
```diff
@@ -0,0 +1,14 @@
...
+
+The model generated by the winery parser (`TServiceTemplate`) is not suitable for this task, as it describes the meta model of the topology. Using the meta model to transform an topology is rather tedious. Goal of the `EffectiveModel` is to provide a more suitable model for given purpose. When categorizing the `TServiceTemplate` as the meta model, the `EffectiveModel` is its instance model.
+
+The model contains all normative Types and the non-normative Docker Types. The different Artifact Types however are not modeled. As they are neither referenced in Capability-, Relationsship- nor Node-Types I decided to give up the distinction between artifact types (for now).
+
+##### Issues:
+- Some Classes might need a better name 
+  - but in general I'd like to keep the names short
+
+![model](effective-model.png)
```
JensMueller2709: the class "DockerEngine" is without connection

hnicke: fixed - DockerEngine is removed from the diagram, as there is no need for the tosca modeler to explicitly model a DockerEngine.
The current approach of adding Docker as supported technology now is in accordance with TOSCA Simple Profile Chap 9 (non-normative Types).

c-mueller: Is the notation `name[0..1]` valid UML? never seen this before? 

It also does not make any sense because a class will have to have the field anyways even if it does not have any value.

c-mueller: Would also propose that the visibilites should be added on the missing fields or should get removed completely

c-mueller: The Size Attributes seem very inconsistent sometimes its Gigabytes and sometimes its megabytes. 
I don#t know if this is specified within TOSCA this way but it looks bad. Especially when looking at ObjectStorage and BlockStorage they both have a similar purpose and one of them is modeled in GB and the other one in MB

hnicke: I looked up how to model multiplicities in attributes, and found a resource that explained it that way. But now, after searching again, I only [found](https://www.uml-diagrams.org/multiplicity.html) several sources that verify `name: String [lower..upper]` is the correct notation. Will change that.

hnicke: I think I'll remove visiblity everywhere. Everything is meant to have a public scope for now.

hnicke: Regarding the inconsistencies in size units: I am fully aware of that. Unfortunately this is the way the TOSCA spec defines the node types. Basically we could transform all values to MB or GB internally, but we somehow lose semantics. I'd leave it as it is.

hnicke:
@jmuell better think now twice; the cost to change the model after its implementation is ten times higher than when we realize now that somethings wrong.

docs/dev/effective-model/effective-model.md
```diff
@@ -0,0 +1,23 @@
+# Effective model
+
+We need a common model that represents the topology of a CSAR. This model shall be used by the plugins to execute the transformation.
+
+The model generated by the winery parser (`TServiceTemplate`) is not suitable for this task, as it describes the meta model of the topology. Using the meta model to transform an topology is rather tedious. Goal of the `EffectiveModel` is to provide a more suitable model for given purpose. When categorizing the `TServiceTemplate` as the meta model, the `EffectiveModel` is its instance model.
+
+The model contains all normative Types and the non-normative Docker Types. The different Artifact Types however are not modeled. As they are neither referenced in Capability-, Relationsship- nor Node-Types I decided to give up the distinction between artifact types (for now).
```
mklopp: Relationship- nor Node-Types

Issue #155, opened 2017-11-07
Remove hardcoded server ip and port in the CLI


nfode:
Currently the CLI uses a hardcoded port and IP. To enable configuration I would suggest to create a properties file in the home dir of the user. This would give a adress like this: `<user-home-dir>/.toscana/cli.properties`. 

To configure the CLI without editing files I would recommend using a subcommand named `config`.

Set the ip and port:
```
toscana config ip "192.168.0.1"
toscana config port "5000"
```
View the settings:
```
toscana config ip 
> 192.168.0.1
```
```
toscana config port 
> 5000
```

In the properties file the values are saved like this:
```
ip=192.168.0.1
port=5000
```
with this notation the file is bash compatible.

c-mueller:
We should keep it together and maybe call it cli endpoint with the value (for example) `http://localhost:8080` this also allows support of HTTPS based instances.

Thi also allows you to set the endpoint to something like `http://192.168.99.100:8080/toscana`you might have a adress like this when running on tomcat

hnicke:
- setting the endpoint via `CLI` is unnecessary
- Instead, i propose following:
  - On `CLI` startup, `CLI` reads endpoint from a config file.
  - Documentation (e.g. man page) must provide information where this file is located
    - proper location for linux: $HOME/.config/toscana/
    - proper location for windows: $HOME/AppData/toscana/ (?)
- sensible default value should be `http://localhost:{port}`, whereas {port} should NOT be `8080`. I propose something like `8083`

We need this external configuration file because resources located in the jar are immutable at runtime.

nfode:
As I mentionend in the PR I do not think setting the url int the CLI is unneccessary, for Example if I had to use our CLI on Windows I do not have any idea how to edit the `.properties` file without extra research.
@hnicke for you as unix user it might seem easy to use `vim ~/.config/toscana/cli.properties` ;)

hnicke:
@nfode Navigate to file with explorer -> right click -> open with -> notepad?

nfode:
@hnicke I think AppData is a hidden folder, so I have to click to about 5 submenus to see the hidden folder...

nfode:
But hopefully I never have to use it on Windows 🤣

hnicke:
A developer should know how to access hidden files. Implement a cli command to set the endpoint if you like -- but we really have other problems than that.

PR #174, opened 2017-11-11
Remove hardcoded API Url from the CLI

orferch:
Implemented after suggestions from #155


cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+public class CliProperties {
+
+    private static String filePath;
+    private static String dataPath = null;
+    private static final String dataPathWin = "/AppData/toscana";
+    private static final String dataPathNix = "/.toscana";
+    private static final String cliProp = "/cli.properties";
+    private static Properties properties = new Properties();
+    private static File file;
+    private static String OS = null;
```
c-mueller: This does not adhear to java code conventions replace this with `operatingSystem` (or simmilar)

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Creates a cli.properties config file
+     */
+    private static void createProperties() {
+        try {
+            FileWriter writer = new FileWriter(file);
+            properties.setProperty("ip","http://localhost");
```
c-mueller: i don't like the idea of splitting this in 3 parts. As i suggested in the original issue one property that includes the whole base url seems to be a way better solution.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+        
+        try {
+            InputStream inputStream = new FileInputStream(file);
+            properties.load(inputStream);
+            inputStream.close();
+        } catch (IOException e) {
+            System.err.println(e.getMessage());
+        }
+        
+        url += properties.getProperty("ip");
```
c-mueller: If we decide to not change this, Use StringBuilder instead

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public static void setupPath() {
+        if (dataPath == null || dataPath.isEmpty()) {
+            // init dataPath to platform dependent value
+            dataPath = System.getProperty("user.home");
+            if (OS == null) {
+                OS = System.getProperty("os.name");
+            }
+            
+            if (OS.contains("Linux") || OS.contains("Mac")) {
```
hnicke: checkout org.opentosca.toscana.core.util.OsUtils. You should reuse that util class for convenience.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+        file = new File(filePath);
+
+        if (!dataDir.exists()) {
+            dataDir.mkdirs();
+            if (!dataDir.exists()) {
+                System.err.println("Failed to create data directory");
+            }
+        }
+
+        if (!new File(filePath).exists()) {
```
hnicke: reuse field `file` insteald of `new File(filePath)`?

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+        url += properties.getProperty("ip");
+        url += ":" + properties.getProperty("port");
+        url += "/" + properties.getProperty("url");
+        return url;
+    }
+
+    /**
+     * Sets the path for Unix or Windows Systems to the cli.properties file
+     */
+    public static void setupPath() {
```
hnicke: In my opinion this is a design flaw. Your class depends on external initialization. This makes working with the class more brittle than it should be. If you decide to keep all the methods static, I suggest you make `setupPath()` private and use an initializer (= static block):
```
static {
    setupPath()
}
```
Then you can also make the field `OS` a local variable.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+            properties.load(inputStream);
+            inputStream.close();
+        } catch (IOException e) {
+            System.err.println(e.getMessage());
+        }
+        
+        url += properties.getProperty("ip");
+        url += ":" + properties.getProperty("port");
+        url += "/" + properties.getProperty("url");
+        return url;
```
hnicke: url can be an empty string here in case of failure to read properties file.
We should provide a fallback value. maybe use the same value as the default values for the properties file.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+    public static String getApiUrl() {
+        String url = "";
+        setupPath();
+        
+        try {
+            InputStream inputStream = new FileInputStream(file);
+            properties.load(inputStream);
+            inputStream.close();
+        } catch (IOException e) {
+            System.err.println(e.getMessage());
```
hnicke: Please provide context in the error message.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+            }
+            
+            if (OS.contains("Linux") || OS.contains("Mac")) {
+                dataPath += dataPathNix;
+                filePath = dataPath + cliProp;
+            } else if (OS.contains("Windows")) {
+                dataPath += dataPathWin;
+                filePath = dataPath + cliProp;
+            } else {
+                dataPath = FileUtils.getTempDirectory() + File.separator + "toscana";
```
hnicke: Who guarantees the current user has write access to `<tmpdir>/toscana`?
Struggled with [same flaw](https://github.com/eclipse/winery/issues/193) in eclipse/winery. Follow the link to see how I suggest to solve the issue.
Countless bugs happen due to incorrect error handling.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,99 @@
...
+                filePath = dataPath + cliProp;
+                System.err.println("fallback value for datadir not defined for this platform. Falling back to tmp dir");
```
hnicke: When logging errors, always provide as much context as you can give. You should print the fallback directory, too! Use String.format() to build log messages to std.out/std.err with elegance:
```
System.err.println(String.format("fallback value for datadir not defined for current platform '%s'. Falling back to '%s'", OS, filepath)
```

docs/dev/cli/cli-commands.md
```diff
@@ -92,6 +92,15 @@ Options explanation:
 ![Class Diagram](img/example_usage.png)
 ***Note:*** Example usage of the CLI for Platform Informations
 
+## Change API Url
```
hnicke: You could use the term endpoint instead of Url

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
hnicke: please use 127.0.0.1 instead of localhost. some hosts (especially some vm images /docker images) are not properly configured and localhost doesn't point to 127.0.0.1, this has often led to extrem irritation.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
hnicke: You should also print the stacktrace.

docs/dev/cli/cli-commands.md
nfode: What is about the idea to give the user commands to set the endpoint? It is a far better solution then navigate manually to the config folder.

nfode:
@orferch what is the current status? 
@hnicke @c-mueller I think if this PR fixes the minor issues we should squash it into the master then create a new branch to refactor it. Is it better to add the wrapper while refactoring or after? I think during the process is better.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,101 @@
...
+        try {
+            InputStream inputStream = new FileInputStream(file);
+            properties.load(inputStream);
+            inputStream.close();
+        } catch (IOException e) {
+            System.err.println("Something went wrong while trying to load the API Endpoint.");
+            e.printStackTrace();
+        }
+
+        url += properties.getProperty(CLI_PROPS_ENDPOINT_KEY);
```
hnicke: Still url can be "", if key is not found. And then it is simply returned to the caller. Callee should report error and use a fallback. User will not get appropriate feedback if he messed up editing the config file.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,101 @@
...
+        return url;
+    }
+
+    /**
```
hnicke: Please check [how to contribute](https://github.com/StuPro-TOSCAna/TOSCAna/blob/master/CONTRIBUTING.md). You need to import the checkstyle and reformat the code. Javadoc comments should then be formatted without leading asterisks, like in the server module.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,101 @@
...
+                final String dataPathNix = "/.toscana";
+                dataPath += dataPathNix;
+                filePath = dataPath + cliProp;
+            } else if (operatingSystem.contains("Windows")) {
+                final String dataPathWin = "/AppData/toscana";
+                dataPath += dataPathWin;
+                filePath = dataPath + cliProp;
+            } else {
+                dataPath = FileUtils.getTempDirectory() + File.separator + "toscana";
+                filePath = dataPath + cliProp;
```
hnicke: Usage of String concatenation to build paths is quite brittle AFAIK. I'd generally use:
```File foo = new File(bar, "foo")```

File: cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java [hunk 15]
```diff
@@ -52,19 +52,17 @@ public String getApiUrl() {
...
             return url;
         } else {
             createProperties();
-            try {
-                throw new IOException("API Endpoint not correctly set, using default value");
-            } catch (IOException e) {
-                System.err.println(e.getMessage());
-            }
+            System.err.println("API Endpoint not correctly set, using default value http://127.0.0.1:8084/");
+            System.exit(2);
```
hnicke: Why do you exit here?

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,109 @@
...
+                file = new File(dataDir, cliProp);
+            } else if (operatingSystem.contains("Windows")) {
+                dataDir = new File(dataPath, "/AppData/toscana");
+                file = new File(dataDir, cliProp);
+            } else {
+                dataPath = FileUtils.getTempDirectory() + File.separator + "toscana";
+                dataDir = new File(dataPath);
+                file = new File(dataDir, cliProp);
+                System.err.println(String.format("fallback value for data directory not defined for current platform '%s'. Falling back to '%s'", operatingSystem, dataDir.getPath()));
+                System.exit(1);
```
hnicke: Whats going on here? Why do you exit the appication? In line `89-91` you build a fallback dir. in line `92-93` you tell the user that supplying a fallback dir is not possible and you exit with failure.
I suggest:
- inform user about fallback dir, as current platform is not supported
- don't exit.

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,109 @@
...
+    private void setupPath() {
+        File dataDir = null;
+        if (dataPath == null || dataPath.isEmpty()) {
+            // init dataPath to platform dependent value
+            dataPath = System.getProperty("user.home");
+            if (operatingSystem == null) {
+                operatingSystem = System.getProperty("os.name");
+            }
+
+            final String cliProp = "/cli.properties";
```
hnicke: remove leading "/"

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,109 @@
...
+            if (operatingSystem.contains("Linux") || operatingSystem.contains("Mac")) {
+                dataDir = new File(dataPath, "/.toscana");
```
hnicke: remove leading "/"

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,109 @@
...
+                file = new File(dataDir, cliProp);
+            } else if (operatingSystem.contains("Windows")) {
+                dataDir = new File(dataPath, "/AppData/toscana");
```
hnicke: remove leading "/'

cli/src/main/java/org/opentosca/toscana/cli/CliProperties.java
```diff
@@ -0,0 +1,109 @@
...
+                System.err.println(String.format("fallback value for data directory not defined for current platform '%s'. Falling back to '%s'", operatingSystem, dataDir.getPath()));
+                System.exit(1);
+            }
+        }
+
+        if (dataDir != null && !dataDir.exists()) {
+            dataDir.mkdirs();
+            if (!dataDir.exists()) {
+                System.err.println("Failed to create data directory");
+                System.exit(1);
```
hnicke: Why do we exit when we can't create the properties file?
I suggest:
- inform user about failure
- use fallback value for endpoint
- don't quit

nfode:
Going to test this now! Please do not add any commits until I am finished.

PR #201, opened 2017-11-14
Use Project Lombok to tackle boilerplate code

hnicke:
I'd like to use [Project Lombok](https://projectlombok.org/) for the implementation of the EffectiveModel.
If you don't like this suggestion please give quick feedback. Otherwise I will begin tomorrow to add Lombok to the EffectiveModel implemenation branch (will save me from writing much boilerplate code).

### DIFF
--- a/docs/adr/0003-lombok.md
+++ b/docs/adr/0003-lombok.md
@@ -1,0 +1,51 @@
+# *Use Project Lombok in order to avoid writing boilerplate code*
+
+**User Story:** 
+
+The implemenation of the EffectiveModel must make use of the builder pattern (many optional parameters required for object construction).
+
+Without the builder pattern, the model has approx. 2.5k loc. With the builder pattern, it grows to approx. 4k loc, due to the boilerplate nature of the pattern.
+
+## Considered Alternatives
+
+* *Use lombok*
+* *Write boilerplate code*
+* *Use kotlin instead of java*
+
+## Decision Outcome
+
+* Chosen Alternative: *Use Lombok*
+* *Comes out best (see below)*
+* *Consequences: Every developer has to install the lombok plugin in his IDE* <!-- optional -->
+
+## Pros and Cons of the Alternatives <!-- optional -->
+
+### *Use lombok*
+
+##### What is lombok
+[Project lombok](https://projectlombok.org/) is an annotation-based framework which strives to eliminate boilerplate code from the codebase.
+It can tackle:
+- [Builder pattern](https://projectlombok.org/features/Builder)
+- [Getters & setters](https://projectlombok.org/features/GetterSetter)
+- many more
+
+#### Pro's and cons
+* `+` *significantly smaller EffectiveModel code base*
+* `+` *better maintainability due to smaller code base*
+* `+` *better readability due to smaller code base*
+* `+` *the `delombok` feature makes it possible to easily rip lombok out of the project if the need arises*
+* `+` *No boilerplate that can get out of sync if class changes*
+* `-` *introduces more "black magic" to the codebase*
+* `-` *developers have to know what each annotation does*
+* `-` *every developer has to install a plugin for his IDE. supported: intellij, eclipse*
+
+### *Write boilerplate code*
+
+* `+` *pure vanilla java*
+* `-` *boilerplate code that might get out of sync when class changes*
+
+### *Use kotlin instead of java*
+
+* `+` *lean language design*
+* `+` *would have similar positive effects than Lombok in terms of maintainability and readability*
+* `-` *team (besides @c-mueller) doesn't know kotlin, and it's much harder to learn than learning Lombok*
--- a/docs/adr/index.md
+++ b/docs/adr/index.md
@@ -7,6 +7,7 @@ This log lists the architectural decisions for *[project name]*.
 - [ADR-0000](0000-use-markdown-architectural-decision-records.md) - Use Markdown Architectural Decision Records
 - [ADR-0001](0001-model-for-transformations.md) - Use custom model for transformations
 - [ADR-0002](0002-general-plugin-architecture.md) - Design a general plugin architecture
+- [ADR-0003](0003-lombok.md) - Use lombok in order to avoid writing boilerplate code
 
 
 0002-general-plugin-architecture.md


docs/adr/0003-lombok.md
```diff
@@ -0,0 +1,54 @@
...
+* `+` *the `delombok` feature makes it possible to easily rip lombok out of the project if the need arises*
+* `+` *No boilerplate that can get out of sync if class changes*
+* `-` *introduces more "black magic" to the codebase*
+* `-` *developers have to know what each annotation does*
+* `-` *every developer has to install a plugin for his IDE. supported: intellij, eclipse*
+
+### *Write boilerplate code*
+
+* `+` *pure vanilla java*
+* `-` *boilerplate*
```
c-mueller: Once is enough

docs/adr/0003-lombok.md
```diff
@@ -0,0 +1,54 @@
...
+* `-` *more boilerplate*
+* `-` *boilerplate code that might get out of sync when class changes*
+* `-` *did I mention boilerplate*
+
+### *Use kotlin instead of java*
+
+* `+` *lean language design*
+* `+` *would have similar positive effects than Lombok in terms of maintainability and readability*
+* `-` *team doesn't know kotlin, and it's much harder to learn than learning Lombok*
```
c-mueller: Well i know Kotlin. :P

docs/adr/0003-lombok.md
```diff
@@ -0,0 +1,54 @@
...
+
+#### Pro's and cons
+* `+` *significantly smaller EffectiveModel code base*
+* `+` *better maintainability due to smaller code base*
+* `+` *better readability due to smaller code base*
+* `+` *the `delombok` feature makes it possible to easily rip lombok out of the project if the need arises*
+* `+` *No boilerplate that can get out of sync if class changes*
+* `-` *introduces more "black magic" to the codebase*
+* `-` *developers have to know what each annotation does*
+* `-` *every developer has to install a plugin for his IDE. supported: intellij, eclipse*
```
c-mueller: This is a huge "problem" (it has to be mentioned in the readme) Please open a issue or update the readme yourself once the plugin is needed

docs/adr/0003-lombok.md
```diff
@@ -0,0 +1,54 @@
+# *Use Project Lombok in order to avoid writing boilerplate code*
+
+**User Story:** 
+
+The implemenation of the EffectiveModel must make use of the builder pattern (many optional parameters required for object construction).
```
c-mueller: I think using this is only needed in the model, all other pojo style classes are not large enough to need any features of the lombok thing

c-mueller:
Well i forgot to write this in the Review comment itself:

Writing stupid boilerplate code should be avoided, therefore using this whenever a BuilderPattern is needed is completely okay for me. But i would keep it at that.

docs/adr/0003-lombok.md
nfode: It has to be added to the Readme but to install a Plugin is not a *huge* problem in my opinion.

docs/adr/0003-lombok.md
nfode: Seems like he does not like boilerplate code? 😄

docs/adr/0003-lombok.md
nfode: I have the plan to learn it...

docs/adr/0003-lombok.md
nfode: Limiting it to the model might be a good idea.

docs/adr/0003-lombok.md
hnicke: I will add it to the readme when adding the dependency to the project.

hnicke:
@c-mueller you said you would like to limit the use of lombok to the builder pattern.
I'd also limit the usage -- but would also like use these features:
- getters/setters. 
    - reduces even more boilerplate code
    - if a more complicated getter/setter is needed, one can simply provide an implementation in this special case, and lombok will honour that.
    - if javadocs are needed, lombok supports a special notation which converts field javadocs to getter/setter javadocs, which should be sufficient
    - only drawback: theres no annotation for generating Optional<Value> getters. But we can simply write our own getters in these cases
- equals/hash
    - another brilliant option to reduce boilerplate, as writing correct equals/hash methods is tedious
    - i don't like how intellij generates these, and would write them manually if we don't use lombok
    - of course, if a more complex equals method is required, it can be supplied manually, and lombok will honour that
    - especially equals/hash methods tend to get out of sync, which leads to mysterious bugs. not so with lombok.

I'm curious what you think.

hnicke:
## Example: Capability.java

@c-mueller @nfode This is what I would want to do.

#### lombok

```java
package org.opentosca.toscana.model.capability;

import java.util.Set;

import org.opentosca.toscana.model.DescribableEntity;
import org.opentosca.toscana.model.node.RootNode;

import lombok.Builder;
import lombok.Data;
import lombok.NonNull;

/**
 Defines a set of data that can be associated with a {@link RootNode} to describe its capability or feature.
 (TOSCAV1.1, p. 82)
 */
@Builder
@Data
public class Capability extends DescribableEntity {

    /**
     Set of Node Class Types that are valid sources of any relationship established to this capability.
     If empty, all Node classes are valid source types.
     (TOSCAV1.1, p. 83)
     */
    private final Set<Class> validSourceTypes;

    @NonNull
    private final Occurence occurence;
}
```

#### vanilla java
```java
package org.opentosca.toscana.model.capability;

import java.util.Set;

import org.opentosca.toscana.model.DescribableEntity;
import org.opentosca.toscana.model.node.RootNode;

import lombok.NonNull;

/**
 Defines a set of data that can be associated with a {@link RootNode} to describe its capability or feature.
 (TOSCAV1.1, p. 82)
 */
public class Capability extends DescribableEntity {

    /**
     Set of Node Class Types that are valid sources of any relationship established to this capability.
     If empty, all Node classes are valid source types.
     (TOSCAV1.1, p. 83)
     */
    private final Set<Class> validSourceTypes;

    @NonNull
    private final Occurence occurence;

    @java.beans.ConstructorProperties( {"validSourceTypes", "occurence"})
    Capability(Set<Class> validSourceTypes, Occurence occurence) {
        this.validSourceTypes = validSourceTypes;
        this.occurence = occurence;
    }

    public static CapabilityBuilder builder() {
        return new CapabilityBuilder();
    }

    public Set<Class> getValidSourceTypes() {
        return this.validSourceTypes;
    }

    @NonNull
    public Occurence getOccurence() {
        return this.occurence;
    }

    public boolean equals(Object o) {
        if (o == this) return true;
        if (!(o instanceof Capability)) return false;
        final Capability other = (Capability) o;
        if (!other.canEqual((Object) this)) return false;
        final Object this$validSourceTypes = this.getValidSourceTypes();
        final Object other$validSourceTypes = other.getValidSourceTypes();
        if (this$validSourceTypes == null ? other$validSourceTypes != null : !this$validSourceTypes.equals(other$validSourceTypes))
            return false;
        final Object this$occurence = this.getOccurence();
        final Object other$occurence = other.getOccurence();
        if (this$occurence == null ? other$occurence != null : !this$occurence.equals(other$occurence)) return false;
        return true;
    }

    public int hashCode() {
        final int PRIME = 59;
        int result = 1;
        final Object $validSourceTypes = this.getValidSourceTypes();
        result = result * PRIME + ($validSourceTypes == null ? 43 : $validSourceTypes.hashCode());
        final Object $occurence = this.getOccurence();
        result = result * PRIME + ($occurence == null ? 43 : $occurence.hashCode());
        return result;
    }

    protected boolean canEqual(Object other) {
        return other instanceof Capability;
    }

    public String toString() {
        return "Capability(validSourceTypes=" + this.getValidSourceTypes() + ", occurence=" + this.getOccurence() + ")";
    }

    public static class CapabilityBuilder {
        private Set<Class> validSourceTypes;
        private Occurence occurence;

        CapabilityBuilder() {
        }

        public Capability.CapabilityBuilder validSourceTypes(Set<Class> validSourceTypes) {
            this.validSourceTypes = validSourceTypes;
            return this;
        }

        public Capability.CapabilityBuilder occurence(Occurence occurence) {
            this.occurence = occurence;
            return this;
        }

        public Capability build() {
            return new Capability(validSourceTypes, occurence);
        }

        public String toString() {
            return "Capability.CapabilityBuilder(validSourceTypes=" + this.validSourceTypes + ", occurence=" + this.occurence + ")";
        }
    }
}
```

nfode:
@hnicke the shortness of the lombok code speaks for itself

Issue #192, opened 2017-11-13
Basic plugin workflow


JensMueller2709:
I thought about a basic plugin workflow. What we get from the core component and what the plugin has to do. Therefore i just draw a simple flow with my first ideas and want to show you to start a discussion about it.

![simple plugin flow - page 1](https://user-images.githubusercontent.com/26904400/32727162-d329a0c6-c87b-11e7-9cce-8aa5a11c680d.png)

I think the image is self explanatory since it is very simple. Neverthess if you have any question, just ask.

nfode:
It is currently the plan that the plugins get a *jgrapht* to work on. No list. What is the job of the OrderManager to just iterate over the List we do not need a extra class. 

But the main thing is that in your basic workflow the whole idea of our transformation lifecycle is missing.

hnicke:
In the last sprint meeting (for sprint 3) @koppor wanted us to stick closer to (propperly modeled) UML when drawing diagrams. You use uml class diagram to describe a workflow here, which is incorrect AFAIK.

In our currently established domain language, the TargetArtifact contains the readme and deployment.sh. (please verify, @nfode ).

I think the plugins can't simply look at the nodetype of a node and read it's attributes. What about unfulfilled requirements of a node? They are meant to be fulfilled by the orchestrator/plugin. This is the reason I'm currently taking much care in designing all the normative capabilities.

nfode:
The is not necessarily called `deployment.sh` but it is definitively include in the target artifact.

PR #206, opened 2017-11-14
[WIP] plugin flow

JensMueller2709:
The purpose of this document is to make a more clear overview of what the plugins have to do.
First discussions you find in Issue #192 

Before it could be merged the flow diagram will be exchanged to UML-standard.


JensMueller2709:
@hnicke i think the plugins have to read the information of the nodes e.g. the properties to fill there yaml or JSONs. I thought the nodes of the graph are nodes of the effective-model-types. And the plugin can handle with these Java-objects. Am I absolutely wrong?

hnicke:
Basically youre right, so far. But the edges of the Graph are the Relationships (dependsOn, connectsTo, etc, each with its own attached operations and attributes). The graph you receive is not necessarily a connected graph -- as the relationships in the TOSCA yaml file might be only implicitly modeled (with the help of requirements). So as a first step, you have to connect the graph. Therefore you may have to add Nodes to the graph. This shall happen in the `prepare` lifecycle phase.

JensMueller2709:
i will delete this branch, because i will create a CloudFoundry specific transformation-flow. Nevertheless, thank you for your input

PR #207, opened 2017-11-14
Add Docs/resource branch category

mmunozba:
Adds a category for branches regarding resource files. (e.g. a CSAR file for testing)


JensMueller2709:
i think this is a good idea for example regarding the LAMP-csar development.

hnicke:
What about using not using 'resource`, but `test`?

nfode:
I don't mind. What does the rest say?

mmunozba:
Sounds good.

c-mueller:
I wont wait for anyone else to approve this. Seems to be clear that it is accepted.

Therefore;
Merged!

PR #216, opened 2017-11-16
Implement Transformation Lifecycle

c-mueller:


server/src/main/java/org/opentosca/toscana/plugins/lifecycle/TransformationLifecycle.java
```diff
@@ -0,0 +1,36 @@
...
+     Has to implement the transform lifecycle operations
+
+     @throws Exception If a exception gets out of the method the transformation will fail
+     */
+    void transform() throws Exception;
+
+    /**
+     Has to implement the clean (cleanup) lifecycle operations
+
+     @throws Exception If a exception gets out of the method the transformation will fail
```
hnicke: Maybe the transformation should not fail when cleanup throws an exception.

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/LifecycleAwarePlugin.java
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     This is a method that allows the Plugin Developer to add additional execution phases
+     <p>
+     Can be implemented by the implementing plugins to add tasks that should be executed after the
+     Validation phase and before the prepare phase
+
+     @param executionTasks Add the Phases in the order that they should executed (first one gets inserted first...)
+     */
+    protected void addPostValidatePhases(List<ExecutionPhase<LifecycleT>> executionTasks) {
```
hnicke: I think adding custom phases is pretty much over-engineered.
Furthermore this mechanism could be counterproductive as may lead to plugins implementing their own phases -- quite the opposite of what we are trying to achieve with streamlining the lifecylce phases.

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/LifecycleAwarePlugin.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /**
+     Performs the executon of the phases in the order that has been defined during the construction of the object
+
+     @param context context for the transformation
+     */
+    @Override
+    public void transform(TransformationContext context) throws Exception {
+        Logger logger = context.getLogger(getClass().getName());
```
hnicke: context.getLogger(getClass()) should be sufficient.

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/LifecycleAwarePlugin.java
```diff
@@ -0,0 +1,167 @@
...
+        //Store current time for time measurement
+        long time = System.currentTimeMillis();
+
+        logger.info("Building Lifecycle interface...");
+        LifecycleT lifecycleInterface = getInstance(context);
+
+        logger.info("This transformation has {} phases.", executionTasks.size());
+        for (int i = 0; i < executionTasks.size(); i++) {
+            ExecutionPhase<LifecycleT> phase = executionTasks.get(i);
+            logger.info("Executing phase \"{}\" ({} of {})", phase.getName(), (i + 1), executionTasks.size());
```
hnicke: if you use `'` instead of `"` you don't have to escape the char.

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/LifecycleAwarePlugin.java
```diff
@@ -0,0 +1,167 @@
...
+        addPostPreparePhases(executionTasks);
+
+        executionTasks.add(new ExecutionPhase<>("transformation", TransformationLifecycle::transform));
+
+        addPostTransformPhases(executionTasks);
+
+        executionTasks.add(new ExecutionPhase<>("cleanup", TransformationLifecycle::cleanup));
+
+        //Make list immutable
+        executionTasks = Collections.unmodifiableList(executionTasks);
```
hnicke: I'd use a local variable to build the list. `executionTasks` can then be made final.

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/TransformationLifecycle.java
```diff
@@ -0,0 +1,36 @@
...
+     This method is intended to check if the  model given in a transformation context
+     (has to be given via a constructor)
+
+     @return True if the vailidation succeeded, false otherwise.
+     Returning false will cause the transformation to fail
+     */
+    boolean checkModel();
+
+    /**
+     Has to implement the prepare lifecycle operations
```
hnicke: What is the plugin supposed to do here? Improve javadoc.

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/TransformationLifecycle.java
```diff
@@ -0,0 +1,36 @@
+package org.opentosca.toscana.plugins.lifecycle;
+
+/**
+ The basic interface to implement the TransformationLifecycle
+ */
+public interface TransformationLifecycle {
+    /**
+     This method is intended to check if the  model given in a transformation context
```
hnicke: Do not start javadoc of a method with `This method..`. Instead, use `Checks if the model...`

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/LifecycleAwarePlugin.java
c-mueller: This is required to mock the transformation context when testing. Mockito does not provide a `anyClass()` matcher

hnicke: doesn't `any()` work?

c-mueller: Nope.
Found a workaround though:

``(Class<?) any(Class.class)` works

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/TransformationLifecycle.java
c-mueller: Added some notes below this line

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/TransformationLifecycle.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     Has to implement the prepare lifecycle operations
+     <p>
+     The Prepare lifecycle phase is described as follows:
+     in this step the plugin processes the graph to transform it in the following step.
+     For example the kubernetes plugin needs to split the graph into container and pods.
+     <p>
+     @throws Exception If a exception gets out of the method the transformation will fail
+     */
+    void prepare() throws Exception;
```
hnicke: Just realized that `throws Exception` is a really bad idea here:
When implementing these methods, it's easy to forget error handling of any checked exception (compiler does not complain..)
I suggest to ditch `throws Exception`. In the executing code one has to make sure to catch all unchecked exceptions.

server/src/test/java/org/opentosca/toscana/plugins/lifecycle/LifecycleAwarePluginTest.java
```diff
@@ -0,0 +1,106 @@
...
+    public void successfulExec() throws Exception {
+        when(lifecycle.checkModel()).thenReturn(true);
+        plugin.transform(context);
+        verify(lifecycle).cleanup();
+    }
+
+    @Test(expected = Exception.class)
+    public void exceptionDuringTransformation() throws Exception {
+        when(lifecycle.checkModel()).thenReturn(true);
+        doThrow(new Exception("Test Exception")).when(lifecycle).transform();
```
hnicke: You are testing the exception mechanism of java..? I'm sure it works without bugs!

server/src/main/java/org/opentosca/toscana/plugins/lifecycle/TransformationLifecycle.java
```diff
@@ -0,0 +1,45 @@
...
+     Has to implement the transform lifecycle operations
+     <p>
+     The Transform Lifecycle phase is described as follows:
+     this is where the real transformation is happening. I.e. the Resources for the target platform get created 
+     such as docker images, manifests...
+     <p>
+     
+     @throws Exception If a exception gets out of the method the transformation will fail
+     */
+    void transform() throws Exception;
```
hnicke: Same as above.

server/src/test/java/org/opentosca/toscana/plugins/lifecycle/LifecycleAwarePluginTest.java
c-mueller: Well this test was intended to check that the implementation of the `transform` method is re-throwing the execption as expected.

The test obviously is senseless, because the exception gets passed through by java.

PR #219, opened 2017-11-16
Bash script generator architecture

nfode:
All plugins need to generate bash scripts so it is better to extract this process from the plugins to avoid duplicated code. This is a first draft a detailed model is following. I am interested what you think about this!


docs/dev/plugins/bash-script-generator/bash-script-generator.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Idea 
+Provide a general so called **BashScriptGenerator** that can build bash scripts out of different components. A component could be the following snippet:
+```bash
+echo "Check if <executable> is available."                                                      
+if ! [ -x "$(command -v <executable>)" ]; then                                                 
+  echo 'Error: <executable> is not installed.' >&2                                             
+  exit 1                                                                                 
+fi     
+```
```
hnicke: Maybe it's better to provide functions in util scripts. When building the scripts, the plugins shall use these util functions, whenever possible. These util scripts need to be in the generated target artifact, of course.
This has the advantage of not generating a boilerplate of code (e.g., check for existence of 10 different executables -- would you rather like to have a) 10x above code or b) 10x `check <executable>` and 1x util script with function ?
A general generator would however still be needed for generating the header (`#!/bin/bash`) and for sourcing the util scripts.

nfode: Forgott the bash functions, I am so stupid! So you want to give the user a general util script or a util script fitted to the targert platform?

hnicke: I'd ship a general util script

hnicke: I'd ship a general util script

hnicke: I suggest we use one general util script which is shipped with the target artifact.
The user does not have to interact with this script - however, the scripts called by the user might depend on this util script.

Platforms might create their own util scripts: e.g. `util-<platformId>.sh,`.

In order to have a nice java interface, every function should be reflected as java function (in the bash script generator).
This java function write's the call to the bash function, including necessary parameters, into the script.

Plugins could extend this ScriptGenerator and add provide custom java functions (with own util script).

nfode: Had the same idea with the java interface. Thx for the input!

hnicke:
You're newest changes reflect my ideas 100%.
Additionally, you could specify 
- how the plugin specific scripts are named and stored within the target artifact
- where the general util script is stored within the target artifact

JensMueller2709:
maybe we need in the scripts some security questions like "do you really want to...? Y/N"

nfode:
I do not think that we need user input. Or should we provide the user the option to fill in properties over the cli after the transformation?

docs/dev/plugins/bash-script-generator/bash-script-generator.md
```diff
@@ -20,10 +23,13 @@ function check_executable () {
...
+
 ```Java
 public void checkExecutable (String executable) {
     list.add("check_executable " + executable);
 }
 ```
 
 A plugin can extend the `BashScriptGenerator` if it has additional functions.
+
+In general the scripts are saved in the `output\scripts\utils\` folder in the target artifact. If a plugin provides is own utils functions the script containing them is saved with the name `utils-<plugin-name>.sh`.
```
hnicke: Maybe use forward slashes to denote paths.
Maybe rename `utils` to `util`.
One would not use the `pluginName`, but the `pluginId` to describe the name of the script.

Please also add the changes to the target artifact convention docs.

JensMueller2709:
I can give you an example what I'm thinking about, not sure if necessary.
If the user want to use a MYSQL Database and want to transform to CloudFoundry-artifact. CF create for this a service, in detail a instance of a specific service with a unique name e.g. "mydb". To connect the application with "mydb" i need the name of the service. 
Ways the plugin could handle this:
- creating a service by itself while the transformation (not recommended because there are different services for each provider and it costs money)
- ask the user while transformation for the service-name
- dont set a value for the service-name and fill in when executing the script   

I still dont know how to handle this, but in case of point three the user have to give an input to the bashscript.

c-mueller:
@JensMueller2709 Why don't we automatically generate a service name from the given environment (Csar Parameters, Platform)?

hnicke:
I heard of some fancy functions capabable of returning randomized strings.

JensMueller2709:
@c-mueller To generate a service-name is not a problem. I see some issues in creating the service-instance automatically. But I didnt have a solution yet, just proposals. 
I keep thinking about it. 
But it seems it is not necessary to implement it in the bash script generator.

nfode:
Updated the document. Now not only a plugin can extend the `BashScriptGenerator` it is possible to extend it for every necessary type of util script. So there can be a `DockerScriptGenerator`, `ZipScriptGenerator`, etc...

nfode:
Reworked after a short talk with @hnicke.

PR #239, opened 2017-11-20
PluginFileAccess createFolder method

nfode:


server/src/main/java/org/opentosca/toscana/core/plugin/PluginFileAccess.java
```diff
@@ -123,10 +123,24 @@ public String getAbsolutePath(String relativePath) throws FileNotFoundException
...
+
      @param relativePath relative (to the transformation content directory) path to a file or directory.
      */
     public void delete(String relativePath) {
         File file = new File(targetDir, relativePath);
         FileUtils.deleteQuietly(file);
     }
+
+    /**
+     Creates folder in the target artifact. Does nothing if folder already exists.
```
hnicke: `target artifact` should be rephrased as `transformation content dirrectory`

hnicke: Add note that method will create folders recursively.

server/src/main/java/org/opentosca/toscana/core/plugin/PluginFileAccess.java
```diff
@@ -123,10 +123,24 @@ public String getAbsolutePath(String relativePath) throws FileNotFoundException
...
+
+     @param relativePath relative (to the transformation content directory) path to a file or directory.
+     @return the created folder.
+     */
+
+    public File createFolder(String relativePath) {
```
hnicke: Do we have to return `File`? I'd return `void`. As you see, I've tried my best in the whole PluginFileAccess to not return a single file.

PR #252, opened 2017-11-24
[WIP] Provide EffectiveGraph

hnicke:
Provide a graph that eases the handling of the EffectiveModel.
The vertices are TOSCA Nodes, the edges the TOSCA Relationships between them.

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -40,6 +40,11 @@
             <artifactId>lombok</artifactId>
             <version>1.16.18</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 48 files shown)


server/pom.xml
```diff
@@ -121,6 +121,11 @@
             <version>4.5.1</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
c-mueller: We should really try to keep the dependencies organized, otherwise this will look horrible at the end.

PR #273, opened 2017-11-28
Update Kubernetes plugin with the TransformationLifecycle

nfode:
Removed everything that is not needed for the "new" Kubernetes plugin. Also, added the TransformationLifecycle.


server/src/main/java/org/opentosca/toscana/plugins/kubernetes/KubernetesPlugin.java
```diff
@@ -39,63 +30,15 @@ private static Platform getPlatformDetails() {
     @Override
     public void transform(TransformationContext context) throws Exception {
```
c-mueller: Please Do not override this method, its already implemented by the lifecycle aware plugin

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/KubernetesPlugin.java
```diff
@@ -39,63 +30,15 @@ private static Platform getPlatformDetails() {
...
-        tNodeTemplateMap.forEach((String s, TNodeTemplate tNodeTemplate) -> {
-            if (tNodeTemplate.getType().getLocalPart().equals("DockerApp")) {
-                template[0] = tNodeTemplate;
-            }
-        });
-        if (template[0] == null) throw new InvalidDockerAppException("DockerApp in given TServiceTemplate not found.");
-        return template[0];
+    @Override
+    protected KubernetesLifecycle getInstance(TransformationContext context) throws Exception {
+        return null;
```
c-mueller: implement the instantiation here using:
```java
return new KubernetesLifecycle(context);
```

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/KubernetesLifecycle.java
```diff
@@ -0,0 +1,36 @@
...
+     @param context because the context is always needed this should never be null
+     It probably gets called by the <code>getInstance</code> method of the LifecycleAwarePlugin
+     */
+    public KubernetesLifecycle(TransformationContext context) throws IOException {
+        super(context);
+    }
+
+    @Override
+    public boolean checkModel() {
+        return false;
```
hnicke: Not implemented methods should always `throw new UnsupportedOperationException()` instead of return `null` or `false`.
Pros are: Not implemented functions sometimes are forgotten; later you excounter an NullPointerexception somewhere, or you wonder why a particular thing doesn't work -- best practice is to always throw this default exception.

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/KubernetesPlugin.java
hnicke: @c-mueller maybe then this needs to become `final`?

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/KubernetesPlugin.java
```diff
@@ -35,67 +26,14 @@ private static Platform getPlatformDetails() {
...
-        String resourceFile = new KubernetesResourceFileCreator().createResourceFileAsString(appName);
-        String resourceFilePath = "/" + appName + "_resource.yaml";
-        fileAccess.access(resourceFilePath).append(resourceFile).close();
-    }
-
-    private void generateReadme(PluginFileAccess fileAccess, String appName) throws IOException {
-        String manual = KubernetesManualCreator.createManual(appName, (appName + "_resource.yaml"));
-        String manualPath = "/Readme.md";
-        fileAccess.access(manualPath).append(manual).close();
+    protected Set<Class<?>> getSupportedNodeTypes() {
```
hnicke: As already discussed this method is not needed anymore (delete it from its superclass)

nfode: Yeah started this before I submitted the other PR so in this context it is still valid ;)

c-mueller: Did merge the other one about a minute ago, just update this branch :)

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/KubernetesPlugin.java
c-mueller: Theoretially, would not force it though.

PR #276, opened 2017-11-28
Add transformation lifecycle to CloudFoundry Plugin

JensMueller2709:


server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CfLifecycle.java
```diff
@@ -0,0 +1,40 @@
+package org.opentosca.toscana.plugins.cloudfoundry;
+
+import java.io.IOException;
+
+import org.opentosca.toscana.core.transformation.TransformationContext;
+import org.opentosca.toscana.plugins.lifecycle.AbstractLifecycle;
+
+import org.slf4j.Logger;
+
+public class CfLifecycle extends AbstractLifecycle {
```
hnicke: Please change to CloudFoundryLifecycle. Someone could think this is the CloudFormationLifecycle.

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CfLifecycle.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private TransformationContext context;
```
hnicke: He deserves to be final.

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CfLifecycle.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public void prepare() {
+
+    }
+
+    @Override
+    public void transform() {
+
```
hnicke: Please either add `throw new UnsupportedOperationException()` or `// noop` in every empty method, depending on context.

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLifecycle.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.opentosca.toscana.core.transformation.TransformationContext;
+import org.opentosca.toscana.plugins.lifecycle.AbstractLifecycle;
+
+import org.slf4j.Logger;
+
+public class CloudFoundryLifecycle extends AbstractLifecycle {
+
+    private final TransformationContext context;
+    private final Logger logger = super.logger;
```
c-mueller: Why do you define a field that has been already defined?

This line is useless. you can always access the logger using `logger` or `this.logger`

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLifecycle.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public CloudFoundryLifecycle(TransformationContext context) throws IOException{
+        super(context);
+        this.context=context;
```
c-mueller: Please reformat your code using `CTRL + ALT + L` (on mac propbably `CMD + ALT + L`) before commiting.

According to the default checkstyle configuration this should be:
`this.context = context;`

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLifecycle.java
hnicke: Actually, the logger of a superclass should not be reused. Reason: The logger has the wrong context (in that case, context is `AbstractLifeCycle.class`, but should be `CloudFoundryLifecycle.class`)

c-mueller: No its not in this case.

The logger gets created using the `getClass()` method wich always is exectued on the class of which the object is a instance of. Therefore this logger is in the CloudFounrdy context

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLifecycle.java
```diff
@@ -0,0 +1,37 @@
+package org.opentosca.toscana.plugins.cloudfoundry;
+
+import java.io.IOException;
+
+import org.opentosca.toscana.core.transformation.TransformationContext;
+import org.opentosca.toscana.plugins.lifecycle.AbstractLifecycle;
+
+public class CloudFoundryLifecycle extends AbstractLifecycle {
+
+    private final TransformationContext context;
```
c-mueller: This is not needed, because the superclass stores the context too.

```java
public abstract class AbstractLifecycle implements TransformationLifecycle {

    /**
     The transformation specific logger that can be used to log to the transformations Log Object
     */
    protected Logger logger;
    /**
     The context on which the transformation should be transformed
     */
    protected TransformationContext context;

    /**
     @param context because the context is always needed this should never be null
     It probably gets called by the <code>getInstance</code> method of the LifecycleAwarePlugin
     */
    public AbstractLifecycle(@NotNull TransformationContext context) throws IOException {
        this.context = context;
        this.logger = context.getLogger(getClass());
```
Looking at this we could propbably make the fields final... Will open a issue for that

PR #282, opened 2017-11-29
Small changes on buildPackCreation

JensMueller2709:
Really small changes.
It is to realize the transformation of the lamp-stack.

### DIFF
--- a/server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryFileCreator.java
+++ b/server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryFileCreator.java
@@ -5,6 +5,10 @@ import java.util.Map;
 
 import org.opentosca.toscana.core.plugin.PluginFileAccess;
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryFileCreator.java
```diff
@@ -56,16 +60,24 @@ private String createService() {
...
         fileAccess.access("/deploy_" + app.getAppName() + ".sh")
-            .append("cf -push " + app.getAppName()).close();
+            .append(deploy).close();
     }
 
-    private void createBuildpackAdditionsFile() throws IOException {
-        String buildPackAdditions = "{\"PHP-EXTENSIONS\":[";
-        for (String bp : app.getBuildpackAdditions()) {
-            buildPackAdditions += "\"" + bp + "\",";
+    //only for PHP
```
nfode: Please add proper `/** */` method comments.

PR #288, opened 2017-11-30
Add CloudFormationLifecycle

mmunozba:
Adds the update for the TransformationLifecycle to the CloudFormation Plugin similar to the other plugins.


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationPlugin.java
```diff
@@ -3,29 +3,32 @@
...
+import org.opentosca.toscana.plugins.lifecycle.LifecycleAwarePlugin;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 @Component
-public class CloudFormationPlugin extends AbstractPlugin {
+public class CloudFormationPlugin extends LifecycleAwarePlugin<CloudFormationLifecycle> {
+    private final static Logger logger = LoggerFactory.getLogger(CloudFormationPlugin.class);
```
c-mueller: This logger seems unecessary, because you currently have nothing to log.

but: Keep it in. didn't whine about it in the other pull requests.

PR #295, opened 2017-11-30
Enum with manifest attributes

JensMueller2709:


server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryManifestAttributes.java
```diff
@@ -0,0 +1,43 @@
+package org.opentosca.toscana.plugins.cloudfoundry;
+
+/**
+ provides attribute-names for the cloud foundry manifest
+ */
+public enum CloudFoundryManifestAttributes {
```
hnicke: Always use singular. `CloudFoundryManifestAttribute`

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryManifestAttributes.java
```diff
@@ -0,0 +1,43 @@
+package org.opentosca.toscana.plugins.cloudfoundry;
+
+/**
+ provides attribute-names for the cloud foundry manifest
+ */
+public enum CloudFoundryManifestAttribute {
```
nfode: The tests are failing because you only renamed the class, please rename the whole file and run the tests locally before you push.

JensMueller2709: ah fuck...really stupid

PR #296, opened 2017-11-30
Implement docker imagetag persistency

hnicke:
Decreases worst case startup time from around 20s to 3s.

- Change root dir for csars from `<datadir>` to `<datadir>/csars` (on disk)
in order to store other stuff in `<datadir>` aswell
- Serializes lastUpdate time and image tag list to disk.
After this patch, will only fetch image tags if not found
on disk or if update interval is exceeded

### DIFF
--- a/server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/BaseImageMapper.java
+++ b/server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/BaseImageMapper.java
@@ -28,7 +28,7 @@ import org.springframework.stereotype.Service;
  This class allows the mapping of OsCapabilities (from a tosca model) to a docker base image.
  When initialized with Spring this class will automatically download the latest tags from the base images
  defined in the <code>DockerBaseImages</code> every 24 hours (by default).
- To update this value please modify the <code>toscana.docker.base-image-mapper.update-interval</code> property (value in
+ To persist this value please modify the <code>toscana.docker.base-image-mapper.persist-interval</code> property (value in
  hours)
  */
 @Service
@@ -40,7 +40,7 @@ public class BaseImageMapper {
     /**
      Administrates the "raw" tag data collected from docker-hub
      */
-    private TagBase tagBase;
+    private TagStorage tagStorage;
 
     /**
      An array containing the base images that this Base Image Mapper tries to map to
@@ -51,14 +51,14 @@ public class BaseImageMapper {
     private MapperEngine engine;
     
     @Autowired
-    public BaseImageMapper(DockerBaseImages[] dockerBaseImages, TagBase tagBase) {
+    public BaseImageMapper(DockerBaseImages[] dockerBaseImages, TagStorage tagStorage) {
         this.baseImages = dockerBaseImages;
-        this.tagBase = tagBase;
-        engine = new MapperEngine(tagBase);
+        this.tagStorage = tagStorage;
+        engine = new MapperEngine(tagStorage);
     }
 
     /**
-     Toggles the first update of the mappings during initialisation in spring
+     Toggles the first persist of the mappings during initialisation in spring
      */
     @PostConstruct
     private void postConstruct() {
@@ -66,10 +66,10 @@ public class BaseImageMapper {
     }
 
     /**
-     Performs the Update of each image and sets the "last update" timestamp at the end
+     Performs the Update of each image and sets the "last persist" timestamp at the end
      */
     private void updateBaseImageMap() {
-        if (tagBase.needsUpdate()) {
+        if (tagStorage.needsUpdate()) {
             logger.info("Updating docker base tags");
             for (DockerBaseImages baseImage : baseImages) {
                 logger.debug("Fetching tags for base image {}", baseImage.name());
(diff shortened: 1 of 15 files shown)


server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/BaseImageMapper.java
```diff
@@ -66,9 +82,39 @@ Stores the update intervall for the base Image mappings (in hours)
...
+    private void deserializeTags() {
+        try {
+            if (imageTagsFile.exists()) {
+                ObjectInputStream ois = new ObjectInputStream(new FileInputStream(imageTagsFile));
+                Object o = ois.readObject();
+                ois.close();
+                if (o instanceof List) {
+                    imageTags = (List<ImageTags>) o;
+                } else {
+                    logger.warn("File '{}' contained wrong data. Deleting file", imageTagsFile);
```
nfode: You say something about deleting files but delete nothing?

hnicke: fixed

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/BaseImageMapper.java
```diff
@@ -66,9 +80,40 @@ Stores the update intervall for the base Image mappings (in hours)
...
+        lastUpdateFile = new File(dataDir, "last_update");
+        deserializeTags();
+
+    }
+
+    private void deserializeTags() {
+        try {
+            if (imageTagsFile.exists()) {
+                ObjectInputStream ois = new ObjectInputStream(new FileInputStream(imageTagsFile));
+                Object o = ois.readObject();
```
c-mueller: Why not try to wrap the list and the last update value in a object and serialize that using Jackson (ObjectMapper)?

hnicke: Good idea, I'll try.

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/api/model/Image.java
```diff
@@ -17,7 +18,7 @@ This class represtents one image (belonging to a tag) of a repository.
  Only the needed fields get stored in a attribute. the rest is put into the AdditionalProperties map
  */
 @JsonInclude(JsonInclude.Include.NON_NULL)
-public class Image {
+public class Image implements Serializable {
```
c-mueller: When using Jackson this can be removed. (I think)

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/TagBase.java
```diff
@@ -0,0 +1,109 @@
...
+import org.opentosca.toscana.plugins.kubernetes.docker.mapper.model.DockerImage;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
```
c-mueller: I would consider this a Repository, because its used to persist data. maybe replace `@Component` with `@Repository` but thats just cosmetics.

server/src/test/java/org/opentosca/toscana/core/BaseJUnitTest.java
```diff
@@ -18,15 +19,22 @@
...
-    public static final File PROJECT_ROOT = new File(System.getProperty("user.dir"));
 
     /**
      Grants disk access. Is reset before every test method.
      */
     @Rule
     public final TemporaryFolder temporaryFolder = new TemporaryFolder(PROJECT_ROOT);
-    // "user.dir" is module root
-    public File tmpdir;
+    private static final String STASIC_TMPDIR = "testsuite-tmpdir-static";
```
c-mueller: This probably means `STATIC_TMPDIR`

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/TagBase.java
```diff
@@ -0,0 +1,109 @@
...
+public class TagBase {
```
c-mueller: Maybe the name `TagStorage` fits better

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/TagBase.java
c-mueller: If it does not cause any issues you could also consider excluding this from the regular tests. Using:
`@Profile("!" + Profiles.EXCLUDE_BASE_IMAGE_MAPPER)`

The BaseImageMapper class does that to prevent loading every time a context is created. This won't change with this update, because the downloaded file is deleted after every test.

PR #300, opened 2017-12-01
Add general node conversion mechanism

hnicke:
Added a general mechanism for node conversion.
Achieved it without a huge switch-case statement, hurray.

Had to add a new csar (empty-topology) in order to test non-parsing relevant features via integration tests, because currently all csars with node templates get rejected (UnsupportedOperationException)


server/src/integration/java/org/opentosca/toscana/core/api/upload/CsarUploadIT.java
```diff
@@ -28,7 +28,7 @@ public void setUp() throws Exception {
     public void testFileUpload() throws Exception {
         System.err.println("Server started!");
 
-        toscanaAPI.uploadCsar("test", TestCsars.CSAR_YAML_VALID_MINIMAL_DOCKER);
+        toscanaAPI.uploadCsar("test", TestCsars.EMPTY_TOPOLOGY);
```
c-mueller: I assume this is a valid csar?

server/src/main/java/org/opentosca/toscana/core/parse/converter/ModelConverter.java
```diff
@@ -0,0 +1,49 @@
...
+        Set<RootNode> nodes = convertNodeTemplates(serviceTemplate.getTopologyTemplate());
+        return new EffectiveModel(nodes);
+    }
+
+    private Set<RootNode> convertNodeTemplates(TTopologyTemplateDefinition topology) throws UnknownNodeTypeException {
+        Map<String, TNodeTemplate> templateMap;
+        if (topology != null) {
+            templateMap = topology.getNodeTemplates();
+        } else {
+            logger.warn("Topology template of service template is empty");
```
c-mueller: I dont like the sound of this error message. There probably is no real better wording but reading it the first time sounds very confusing

server/src/main/java/org/opentosca/toscana/core/parse/converter/NodeConverter.java
```diff
@@ -0,0 +1,149 @@
...
+import org.eclipse.winery.model.tosca.yaml.TNodeTemplate;
+
+public class NodeConverter {
+
+    public static final String TOSCA_PREFIX = "tosca.nodes.";
+
+    private Map<String, BiFunction<String, TNodeTemplate, RootNode>> conversionMap = new HashMap<>();
+
+    NodeConverter() {
+        addRule("Compute", this::toCompute);
```
c-mueller: A short comment Explaining the meaning of these parameters would be good

server/src/integration/java/org/opentosca/toscana/core/parse/converter/NodeConverterIT.java
```diff
@@ -0,0 +1,72 @@
...
+            logger.info("Testing conversion of type '{}'", nodeType);
+            TNodeTemplate template = new TNodeTemplate();
+            template.setType(new QName(nodeType));
+            try {
+                converter.convert("name", template);
+                logger.info("Node Type '{}': known", nodeType);
+            } catch (UnsupportedOperationException e) {
+                logger.info("Node Type '{}': known (not yet supported)", nodeType);
+            } finally {
+                System.out.println();
```
c-mueller: is this needed?

server/src/main/java/org/opentosca/toscana/core/parse/CsarParseServiceImpl.java
```diff
@@ -6,18 +6,20 @@
...
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 @Service
 public class CsarParseServiceImpl implements CsarParseService {
 
-    private Logger logger;
+    private Logger logger = LoggerFactory.getLogger(CsarParseService.class);
```
c-mueller: I think we had the intention of sending the logs to the client if the parsing failed!

server/src/integration/java/org/opentosca/toscana/core/api/upload/CsarUploadIT.java
hnicke: Yeah, will rename it.

server/src/integration/java/org/opentosca/toscana/core/parse/converter/NodeConverterIT.java
hnicke: Yes, it increases the test output readability

server/src/main/java/org/opentosca/toscana/core/parse/CsarParseServiceImpl.java
hnicke: We will. I will add documentation that explains why this happens.
- as you see in the method beneath, the logger is assigned to the csar's logger.
- this LoggerFactory.getLogger() assignment is actually only needed for testing purposes:
It's possible to call the (also public) method `parse(File)` (which is used internally in `parse(Csar)` for testing purposes. In this case, `logger` would be null.

server/src/main/java/org/opentosca/toscana/core/parse/converter/ModelConverter.java
hnicke: I will try to reword it to

server/src/main/java/org/opentosca/toscana/core/parse/converter/NodeConverter.java
hnicke: Fixed

PR #306, opened 2017-12-01
Add missing os capability to the compute node

nfode:


server/src/main/java/org/opentosca/toscana/model/node/Compute.java
```diff
@@ -96,10 +102,10 @@ protected Compute(String privateAddress,
...
      */
     public static ComputeBuilder builder(String nodeName,
-                                  AdminEndpointCapability adminEndpoint,
-                                  ScalableCapability scalable,
-                                  BindableCapability binding,
-                                  Requirement<AttachmentCapability, BlockStorage, AttachesTo> localStorage) {
+                                         AdminEndpointCapability adminEndpoint,
+                                         ScalableCapability scalable,
+                                         BindableCapability binding,
+                                         Requirement<AttachmentCapability, BlockStorage, AttachesTo> localStorage) {
```
hnicke: You forgot to add the os capability to the builder.

PR #312, opened 2017-12-01
Add Lamp-test-model

JensMueller2709:


server/src/test/java/org/opentosca/toscana/plugins/testdata/LampApp.java
```diff
@@ -0,0 +1,159 @@
...
+import org.opentosca.toscana.model.node.MysqlDbms;
+import org.opentosca.toscana.model.node.RootNode;
+import org.opentosca.toscana.model.node.WebApplication;
+import org.opentosca.toscana.model.operation.Operation;
+import org.opentosca.toscana.model.operation.OperationVariable;
+import org.opentosca.toscana.model.operation.StandardLifecycle;
+import org.opentosca.toscana.model.relation.AttachesTo;
+import org.opentosca.toscana.model.relation.HostedOn;
+
+public class LampApp {
```
hnicke: This class needs to get refactored. Please don't use these comment dividers between logical chunks. There is a quite nice language feature in java called `methods`. Check it out :1st_place_medal: 
These so called `methods` will later be extremely useful: Correctly used, they offer great reusability value

PR #313, opened 2017-12-01
Reorganize Test Folder structure

### LINKED ISSUE #297: How to seperate Unit and Integration tests
We are currently in a state where we get some integration tests. therefore we should investigate a solution to seperate them from regular unit tests.

Once we found a solution, i will add a ADR when realizing the chosen option

Possible Approaches:
- Use JUnit [Categories](https://github.com/junit-team/junit4/wiki/categories)
	- `+` Supported by Maven and IntelliJ (Categories to execute can be selected)
	- `+` Quite easy to use/setup
	- `-` Both test categories feature the same `source` and `resources` folder
- Use Seperated `integration` and `test` folders [See here](https://www.petrikainulainen.net/programming/maven/integration-testing-with-maven/)
	- `+` Clear differentiation between integration and unit tests
	- `-` Integration tests have to contain a specific string (e.g. the files have to end with `IT.java`)
- Put the integration tests in a seperate module
	- `+` Clear differentiation between integration and unit tests (even clearer than with a IT folder)
	- `-` Possible problems when computing the coverage from integration tests

I personally would prefer the first option (JUnit Categories). Because it is still possible to seperate the tests by putting them in different packages. (As long as there is no issue with visibilies)

c-mueller:
Once done this PR resolves #297.

The following tasks still have to be done:
- [x] Add ADR for the chosen folder Structure
- [x] Create a small note on how to use the modified environment
- [x] Check if the integration resource folder is loaded
- [x] Collect Coverage Data from integration tests
- [x] Provide default intelij execution Profiles
- [x] Update Travis.yml
- [x] Update Jenkinsfile

Merging this results in the following consequence:
You have to add the Resource Directory of the integration tests manually using:
unbenannt

Right Click the Directory -> "Mark Directory as" -> "Test Resource Root"

Please also note that the integration folder and the test folder still result in the same classpath when executing. Class Names and paths in the resource folder have to be unique when combining the two directories.


docs/adr/0011-put-integration-tests-in-seperate-folder.md
```diff
@@ -0,0 +1,44 @@
...
+* Use JUnit Categories and the same `test` directory to seperate between Integration and Unit Tests
+* Store Integration and Unit Tests in seperate directories and use JUnit Categories to categorize them
+* Store Integration and Unit Tests in seperate directories and use a required Suffix on Integration Test Classes
+* Migrate project to gradle
+
+## Decision Outcome
+
+* Chosen Alternative: Store Integration and Unit Tests in seperate directories and use JUnit Categories to categorize them
+* While still Clearly seperating Unit and Integration tests this method still allows a fine grained categorisation of tests.
+* Consequences: The resource directory of the `ìntegration` folder has to be manually set within the IDE
```
hnicke: I think this is not a consequence: I didn't have to do it, was done automatically

docs/dev/server-testing.md
```diff
@@ -0,0 +1,91 @@
+# Testing Guidelines for the Server module
```
hnicke: Please correct spelling and grammar mistakes in this file. I've found a handful.

server/src/integration/java/org/opentosca/toscana/core/BaseJUnitIntegrationTest.java
```diff
@@ -0,0 +1,9 @@
+package org.opentosca.toscana.core;
+
+import org.opentosca.toscana.IntegrationTest;
+
+import org.junit.experimental.categories.Category;
+
+@Category(IntegrationTest.class)
+public abstract class BaseJUnitIntegrationTest extends BaseJUnitTest {
```
hnicke: What the heck is a BaseJUnitIntegrationTest? :grin:
A test is either an integration or an unit test; never a UnitIntegrationTest. I am confused. Can you explain?

server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/docker/image/DockerImageBuilderIT.java
```diff
@@ -44,12 +47,12 @@ public void testBuildShaImage() throws Exception {
     }
 
     private boolean isDockerAvailable() {
```
hnicke: Do we still need this at all?

server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/MapperTagLoaderIT.java
```diff
@@ -19,10 +21,11 @@
 import static org.opentosca.toscana.model.capability.OsCapability.builder;
 
 @ActiveProfiles( {"bimloader"})
```
hnicke: Do we still need the profile?

server/src/test/java/org/opentosca/toscana/UnitTest.java
```diff
@@ -0,0 +1,4 @@
+package org.opentosca.toscana;
+
+public interface UnitTest {
```
hnicke: Why is this required?

server/src/integration/java/org/opentosca/toscana/IntegrationTest.java
```diff
@@ -0,0 +1,4 @@
+package org.opentosca.toscana;
+
+public interface IntegrationTest {
```
hnicke: Why is this required?

c-mueller: This interface and the UnitTest one represents a [Test category in JUnit](https://github.com/junit-team/junit4/wiki/categories)

server/src/integration/java/org/opentosca/toscana/core/BaseJUnitIntegrationTest.java
c-mueller: Well it represents a Integration test that does not need a spring context. Didn't think of a better name. But maybe `BaseIntegrationTest` is better

server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/docker/image/DockerImageBuilderIT.java
c-mueller: I would keep it. just to be save. When trying to run the integration tests with no docker daemon available this test will fail.

server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/MapperTagLoaderIT.java
c-mueller: Yes.

server/src/integration/java/org/opentosca/toscana/core/BaseJUnitIntegrationTest.java
hnicke: Yes! Maybe also rename BaseJunitTest to BaseUnitTest.

docs/adr/0011-put-integration-tests-in-seperate-folder.md
c-mueller: Lets say it might cause issues. Because as I said already i had to set it manually even after restarting the IDE

server/src/integration/java/org/opentosca/toscana/core/BaseJUnitIntegrationTest.java
hnicke: Yeah technically BaseJUnitIntegrationTest is not a contradiction; but then `JUnit` information is quite unnecessary..

hnicke:
Good work!

PR #315, opened 2017-12-01
Feature/mdl/visitor enhanced

hnicke:
## Problem
When implementing special visitors which should only operate on specific entity types, all `visit` methods for types other than these specific types had to be implemented as noop-method (else they would throw Exception).

## Solution
Provide strict and non-strict visitors.
Strict visitors: When non-overriden methods get invoked, throw UnsupportedTypeException.
non-strict visitors: When non-overriden methods get invoked, do nothing.

## After this patch got applied
Plugin devs need to double check that they use the appropriate visitor type for each use case.


hnicke:
Going to merge this now.
@mklopp Please make sure you are aware of these changes.

PR #319, opened 2017-12-02
Add environmentCheck for bashScripts

JensMueller2709:


server/src/main/java/org/opentosca/toscana/plugins/scripts/EnvirnonmentCheck.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ checks if a given environment exists 
+ */
+public class EnvirnonmentCheck {
+    
+    protected EnvirnonmentCheck() {
+    }
+    
+    public String checkEnvironment(String command){
+        String check = " echo \"Check if \"" + command + "\" is available.\"\n" +
```
c-mueller: Please dont copy scripts from the resource directory into source code files. Consider loading it using 
```java
InputStream input = getClass().getClassLoader().gerResourceAsStream("plugins/scripts/util/environment-check.sh");
String script = IOUtils.toString(input); //Did not test this part. it supposed to load the whhole file in a string
```

Even with that said. you probably don't need to copy this anyways because the `AbstractTransformationLifecycle` does copy the util folder (the resource folder in witch this script is located)

if i am not mistaking you should be able to call 
```bash
check <my-programm>
```
within a script. Maybe @nfode can confirm this.

PR #326, opened 2017-12-04
Add CloudFormation-Builder dependency

mmunozba:
Add CloudFormation-Builder dependency.


nfode:
If the library has bugs or misses features but is in general useful consider forking and enhancing it.

server/src/test/java/org/opentosca/toscana/plugins/cloudformation/CFBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+import com.scaleset.cfbuilder.core.Module;
+import com.scaleset.cfbuilder.core.Template;
+import com.scaleset.cfbuilder.ec2.Instance;
+import com.scaleset.cfbuilder.ec2.SecurityGroup;
+import com.scaleset.cfbuilder.ec2.SecurityGroupIngress;
+import org.junit.Test;
+
+import static org.junit.Assert.assertNotNull;
+
+public class CFBuilderTest extends Module {
```
c-mueller: Please rename this class!

If you want to use abbreviations please consider using CFN instead of CF to prevent confusion with CloudFoundry!

server/src/test/java/org/opentosca/toscana/plugins/cloudformation/CFBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+//        assertNotNull(template);
+//        System.err.println(template.toString());
+    }
+
+    class TestModule extends Module {
+
+        public void build() throws Exception {
+            
+            Object keyName = option("KeyName").orElseGet(
+                () -> strParam("KeyName").type("AWS::EC2::KeyPair::KeyName").description("Name of an existing EC2 KeyPair to enable SSH access to the instances").constraintDescription("must be the name of an existing EC2 KeyPair."));
```
c-mueller: This line is way too long. consider putting the Hardcoded strings in static constant fields

server/pom.xml
```diff
@@ -43,6 +43,12 @@
             <artifactId>kubernetes-client</artifactId>
             <version>2.6.2</version>
         </dependency>
+        <!--Cloudformation plugin-->
```
c-mueller: The base image mapper is still part of the K8s plugin. please consider putting your dependencies above the k8s plugin to keep it structured

server/src/test/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationBuilderTest.java
```diff
@@ -0,0 +1,82 @@
...
+import com.scaleset.cfbuilder.ec2.SecurityGroup;
+import com.scaleset.cfbuilder.ec2.SecurityGroupIngress;
+import org.junit.Test;
+
+import static org.junit.Assert.assertNotNull;
+
+public class CloudFormationBuilderTest extends Module {
+    
+    @Test
+    public void simpleTest() throws Exception {
```
c-mueller: What does this test do?

Please consider reading [This](https://github.com/StuPro-TOSCAna/TOSCAna/blob/master/docs/dev/server-testing.md). Just to quote the important part:

>The name of a test case (method) within a class should express the purpose of the test in camel-case as concise as possible.

mmunozba:
Since we decided to fork the cloudformation-builder [here](https://github.com/StuPro-TOSCAna/cloudformation-builder), I think it is more appropriate to move the CloudFormationBuilderTest to that repository. This is due to the purpose of said test being only to test the model, without testing or even requiring anything from the main repository. In the future there will be a need for tests in the main repository, especially end-to-end tests and tests regarding the visitor implementation of CloudFormation, but the test in this PR is not one of those.

Therefore the purpose of this PR is only to add the dependency for our cloudformation-builder.

PR #328, opened 2017-12-04
Handling input values

JensMueller2709:


server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/visitors/CloudFoundryNodeVisitor.java
```diff
@@ -95,9 +105,12 @@ public void handleStandardLifecycle(RootNode node) {
 
             // add inputs to environment list
             for (OperationVariable input : operation.getInputs()) {
-                myApp.addEnvironmentVariables(input.getKey());
+                if (input.getValue().isPresent()) {
```
hnicke: Please refactor: I see code duplication (81-87 = 108-113).

PR #339, opened 2017-12-07
Add CloudFormationNodeVisitor for LAMP-stack

mmunozba:


mklopp:
Database Node visitor not finished yet

mmunozba:
This PR is ready for review.

server/src/test/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationPluginTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+public class CloudFormationPluginTest extends BaseUnitTest {
```
c-mueller: Please consider replacing this with a ModelTransformIntegration test like [this one](https://github.com/StuPro-TOSCAna/TOSCAna/blob/master/server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/KubernetesLampIT.java). Most of the code can probaly be adopted.

Would not consider it a problem tough

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
```diff
@@ -0,0 +1,277 @@
...
+            handleOperation(configure, serverName, CONFIG_CONFIGURE);
+        }
+    }
+
+    private String checkOrDefault(Optional<String> optional, String def) {
+        return optional.isPresent() ? optional.get() : def;
+    }
+
+    private Integer checkOrDefault(Optional<Integer> optional, Integer def) {
+        return optional.isPresent() ? optional.get() : def;
```
c-mueller: you could consider using `optional.orElse(def)`.

This also applies to the other occrances

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationLifecycle.java
```diff
@@ -1,33 +1,66 @@
...
+        } catch (Exception e) {
+            logger.error("Transformation to CloudFormation unsuccessful. Please check the StackTrace for more Info.");
+            e.printStackTrace();
+        }
     }
 
     @Override
     public void cleanup() {
-        throw new UnsupportedOperationException();
+//        TODO implement cleanup
```
c-mueller: Hust put NOOP there. I don't think the cleanup phase is specifically implemented by itself.

you might add something here if its needed.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationModule.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     Get a ref to the KeyName of this template
+     */
+    public Object getKeyNameVar() {
+        return this.keyNameVar;
+    }
+
+    private Fn getUserDataFn(String resource, String configsets) {
+        return Fn.fnDelimiter("Join", "",
```
c-mueller: This looks bad. is it possible to put his into a constant?

if not just leave it as is

server/src/test/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationPluginTest.java
mklopp: I added the transform test but keeping this for now

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationModule.java
mklopp: we'll look into this

mmunozba: Instead of putting the whole thing into a constant, I split the constant part and the part that needs the method parameters up in order to make the method more clear.

mmunozba:
I think this PR is now ready to be merged.

c-mueller:
Does anyone else consider a review?


If not i would try to interactively rebase this down to 2 - 8 commits to give both of you some of the additions and deletions. 

For the next terms please create branches for smalller features and open a pull request that has been done by one person. This makes squashing fairly way easier.

However you need some kind of code base to do so and if the codebase isn't there (like before) you cant do this.

PR #341, opened 2017-12-07
Refactor effective model

hnicke:
While building the converter I had to change some aspects of the model, e.g.:
- Make it more convenient to construct entities - Entities which can be constructed without any parameter are not required anymore as parameter in the builder() method
  - the test model instances (minimal-docker and lamp) are adapted aswell
- No Builder anymore required as parameter for other builders
- Builder classes follow the inheritance topology, too (needed for the converter visitors)
- minor fixes here and there
- code reformat


server/src/main/java/org/opentosca/toscana/model/node/WebServer.java
```diff
@@ -76,4 +69,134 @@ public static WebServerBuilder builder(String nodeName,
     public void accept(NodeVisitor v) {
         v.visit(this);
     }
+
```
c-mueller: Is this delombok intentional?

hnicke: no - will fix that - thx.

PR #343, opened 2017-12-09
DockerFileBuilder command sorter

nfode:


server/src/test/java/org/opentosca/toscana/plugins/kubernetes/docker/dockerfile/builder/DockerfileBuilderTest.java
```diff
@@ -38,4 +39,23 @@ public void testBuildUncompressed() throws Exception {
         System.out.println(s);
         validate(s, UNCOMPRESSED_PATH);
     }
+
+    @Test
+    public void orderedCopyAndEnvTest() {
+        String testString1 = "teststringdererste";
```
c-mueller: Maybe use english here (not a real issue though)

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/dockerfile/builder/DockerfileBuilder.java
```diff
@@ -189,6 +212,16 @@ private void writeTo(Writer out) {
         }
     }
 
+    /**
+     Add all EnvCommands at the beginning of the entries list
+     */
+    private void addEnvCommandsToEntries() {
+        for (EnvCommand envCommand : envCommands) {
```
c-mueller: you could also use [`entries.addAll(1, envCommands)`](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#addAll-int-java.util.Collection-)

PR #351, opened 2017-12-12
Kubernetes Service and DNS overview

nfode:


docs/dev/plugins/kubernetes/service.md
```diff
@@ -0,0 +1,22 @@
...
+| **ClusterIP** | Exposes the service on a cluster internal IP. That means a Service can only reached from within the cluster | It can be used to expose f.ex. the IP from a Service that points to a Pod containing a MySql Database. Other Pods in the Cluster can now connect to the Database without knowing the exact IP adress of the MySql Pod they only need to know the service IP.         |
+| **NodePort**|Exposes the service on each Node's IP at a static port. Creating a NodePort automatically creates a **ClusterIP** to which the NodePort service will route. The **NodePort** service adress from outside is **\<NodeIp>:\<NodePort>**| This can be used to adress a service on a given from outside the cluster. |
+|**LoadBalancer**|Exposes the service externally using a cloud provider's load balancer.| This can be used to expose application in Pods that are distributed over several nodes. It can be used to transform the **tosca.nodes.LoadBalancer**. |
+|**ExternalName**|Maps the service to the **externalName** field.| Can be used to expose a service with a readable name, f.ex: **foo.bar.example.com**|
```
c-mueller: I Don't think we need this. The ExternalName service allows "linking" to a external service. like a PaaS database.

See here:
https://akomljen.com/kubernetes-tips-part-1/

c-mueller: ExternalIP is probably more important. it will require user interaction though.

https://kubernetes.io/docs/concepts/services-networking/service/#external-ips

nfode: I dont get why exactly we need ExternalIP

c-mueller: External IP allows the exposure of a service on a specific IP of the cluster.

Lets say: Kubernetes hosts a webserver and has the node IPs 10.0.0.1 and 10.0.0.2. (external)

if you want to expose the webserver you could use NodePort, but you have to know on what node the server is running, The port number also does not match the one used by the service. but you could also use External IP (when knowing the IP of a node, in the example case: 10.0.0.1)
When doing so the webserver is accessible externally by calling the IP 10.0.0.1 on the port specified in the service definition. 

You have to use this if you want to expose a Ingress LoadBalancer (Like Nginix, traefik...) for example. You can then point DNS records at the given IP

nfode: So for ExternalIP I dont need to know on which node my web app is running I only need to know the IP of some random node in the cluster?

c-mueller: you need to know the ip specified in the external ip section. kubernetes will handle internal routing if its hosted on a different node

c-mueller: it (probably) is a ip of any node in the cluster (does not have to be the one running the webserver)

PR #333, opened 2017-12-05
Implement LAMP Stack transformation

c-mueller:
This PR will resolve #332 , #348 and #337 .

What does it?
It takes the given Topology, extracts Stacks and builds Docker Images to form a result

PR #355, opened 2017-12-14
Create general node sorter by creating node stacks and sorting them

nfode:
This is a rough draft with ideas I had about sorting the nodes in the EffectiveModel to find the order in which they should be deployed. 

**Important:** This draft and the following implementation has or should have impact on every plugin so ideas and feedback by every plugin team/team member are needed and highly appreciated.


c-mueller:
As i already mentioned in our recent Mumble meeting. I think the first option is the way to go, at least for Kubernetes. Because we want to put every Top-Level node (in this case WebApp1 or WebApp2) in seperate containers. THis is also the approach the first Transformation Approach (PR #333) uses.


If another plugin needs a different approach, we could consider implementing different "Splitting Strategies" depending on the plugin.

mmunozba:
At first it seemed to me like **option 1** wouldn't make any sense for the AWS plugin, but after giving it some thought, I think we could definitely make it work.

This was my initial reaction:
>In our current transformation mapping, each Compute node gets a corresponding EC2 instance. 
In the given example it would mean that instead of having one EC2 instance with both Webapps we would have those Webapps on different EC2 instances. A transformation where both end up on the same instance would be a lot closer to the original template. This option would probably also require additional steps during the transformation if e.g. the two Webapps have a conntectsTo-relationship between them since they would be hosted on different EC2 Instances.

In the "cons" section of option one this issue seems to be addressed:
>"Duplicated" nodes in different stacks (can be avoided by checking other stacks for the same node)

So, in order to make this option with node stacks work, every time we visit a new node we would have to check in the cloudformation model whether the node already exists. In our current cloudformation model this shouldn't be an issue. 
Handling the relationships should not be a problem either, all we would have to make sure is that the names of the resources in our cloudformation model match the names in the original template.
One thing I'm not so sure about is SecurityGroups I'd have to have a closer look at them, maybe @mklopp has some input on this.

**Option 2** seems like it could work too. Compared to option 1 it would eliminate the need for duplicate node checks. You say that it could get really "complex", could you elaborate a bit on that? Obviously handling substacks would require additional work, but for me this option seems more intuitive and I feel like it would be easier to deal with for transformations to AWS

I don't really like **option 3**. I think building a graph with the stacks and figuring out how to iterate through said graph can be quite difficult, especially considering (like you say in the "cons") multiple different connections between stacks.


These are my thoughts on the given options, I'm not sure if I could find a better fitting solution for AWS.
I'd like @mklopp to look at this since he has been working with CloudFormation longer than me, so he probably has additional input on this matter.

nfode:
In **Option 2** there can be a stack with several substacks the substacks can have also substacks, this is what I have meant with it can get quite complex. With **Option 1** you just have to "save" that you converted this node to a EC2 or whatever instance. Then if you visit the same node again in another stack you can check for that node and see ah I already handled this one and just put everything above on the same EC2 instance. But as @c-mueller mentioned maybe we add different "node stack type options" that are created outside the plugins. 

Sorting in **Option 2** can also get quite complex. Think of sorting deeply nested substacks.

I start with the implementation of **Option 1** since the kubernetes team needs it any way. Would be interesting to hear what the CloudFoundry team has to add to this discussion. It would be nice to bring everything down to one option.

mklopp:
I agree with @mmunozba. I don't think the redundancy of the server/apache should be a problem. If I understand it correctly a database hosted on the server should then result in another stack where we can just cut the server. As of right now we create SecurityGroups corresponding to a node/resource and because SecurityGroups or similar things are contained in TOSCA policies, which we are currently not supporting, we shouldn't have a problem.

If the kubernetes team needs the Option 1 anyways this needs to be implemented and at worst we just use it as a top-down strategy.

JensMueller2709:
For the CloudFoundry-plugin **option1** seems to be the best solution too. Cloud Foundry tries to make given applications cloud-ready and therefore every application will launched in an own container. Thats similiar with the **option1** I think. 
Maybe @orferch has some further comments.

PR #357, opened 2017-12-15
Create common LAMP-Model

mklopp:
Modify the LAMP model and the Lamp-Csar so it can be used by all plugins.


mklopp:
I still don't know what to put in the `database_host` variable because I think every plugin needs something different.

PR #358, opened 2017-12-15
Fix Kubernetes plugin not runnning a transformation

c-mueller:
while trying to integrate a docker registry i encountered an issue with the transformation process. these "minor" problems have been fixed in this PR. (They are minor in terms of effort, but without these fixes nothing works)


c-mueller:
Going to merge this directly, without a review, because it is really minor i just decided to put it in a seperate branch because i have some issues with the Docker daemon running on my machne. I therefore had to test the thing using the CI

PR #363, opened 2017-12-15
Implement Building mechanism for Docker Registry

c-mueller:


server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/docker/image/PushingImageBuilderIT.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private DockerRegistryCredentials creds;
+
+    private String tag = "sha-test-" + Long.toString(System.currentTimeMillis(), 16);
+
+    private String registryid;
+
+    @Override
+    public void init() throws Exception {
+//        Assume.assumeNotNull(
```
nfode: Please delete this if this is no needed?

c-mueller: I would keep it in, because that is the part used to run the test against a different registry.

c-mueller:
I will merge this into master now, if the CI passes. 

Using this will envolve some fixes in the common yaml model. But that will be part of a seperate PR

PR #372, opened 2017-12-18
Bugfixes CloudFoundry

JensMueller2709:
- [x] empty variables
- [x] wrong file directions


server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryFileCreator.java
```diff
@@ -81,13 +97,16 @@ private void createEnvironmentVariables() throws IOException {
...
     private void createService() throws IOException {
-        ArrayList<String> services = new ArrayList<>();
-        services.add(String.format("  %s:", SERVICE.getName()));
-        for (Map.Entry<String, CloudFoundryServiceType> service : app.getServices().entrySet()) {
-            services.add(String.format("    - %s", service.getKey()));
-        }
-        for (String service : services) {
-            fileAccess.access(MANIFEST).appendln(service).close();
+        Map<String, CloudFoundryServiceType> appServices = app.getServices();
+        if (!appServices.isEmpty()) {
```
orferch: maybe a nullcheck is needed here? if app.getServices() is null, nullerror will be thrown

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryFileCreator.java
```diff
@@ -131,18 +150,25 @@ private void createDeployScript() throws IOException {
...
         JSONObject buildPackAdditionsJson = new JSONObject();
         JSONArray buildPacks = new JSONArray();
         for (String buildPack : app.getBuildpackAdditions()) {
             buildPacks.put(buildPack);
         }
         buildPackAdditionsJson.put(BUILDPACK_OBJECT_PHP, buildPacks);
-        fileAccess.access(BUILDPACK_FILEPATH_PHP).append(buildPackAdditionsJson.toString(4)).close();
+        String path;
+        if (app.getMainApplicationPath() != null) {
+            path = app.getMainApplicationPath() + "/" + BUILDPACK_FILEPATH_PHP;
```
orferch: StringBuilder would be a better fit here

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/visitors/CloudFoundryNodeVisitor.java
```diff
@@ -87,12 +87,16 @@ private void handleStandardLifecycle(RootNode node) {
         for (Operation operation : node.getStandardLifecycle().getOperations()) {
             // artifact path
             if (operation.getArtifact().isPresent()) {
-                myApp.addFilePath(operation.getArtifact().get().getFilePath());
```
orferch: refactor can be done here with both if clauses

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryFileCreator.java
```diff
@@ -55,24 +59,36 @@ public void createFiles() throws IOException, JSONException {
...
-        fileAccess.access(MANIFEST).appendln(manifestHead).close();
+        fileAccess.access(MANIFEST_PATH).appendln(manifestHead).close();
     }
 
-    private void createEnvironmentVariables() throws IOException {
-        ArrayList<String> environmentVariables = new ArrayList<>();
-        environmentVariables.add(String.format("  %s:", ENVIRONMENT.getName()));
-        for (Map.Entry<String, String> entry : app.getEnvironmentVariables().entrySet()) {
-            environmentVariables.add(String.format("    %s: %s", entry.getKey(), entry.getValue()));
+    private void addPathToMainApplication() throws IOException {
```
orferch: could call it addPathtoApplication() and delete main word everywhere else

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryFileCreator.java
JensMueller2709: I dont think so, because the app.getServices() cant be null. It returns always a list as you can see in the CloudFoundryApplication class. 
I think it is sufficient to check if the list is empty

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/visitors/CloudFoundryNodeVisitor.java
JensMueller2709: I am not sure what exactly you mean, but I add some commands to an external method

nfode:
Please update this branch with the master.

PR #376, opened 2017-12-19
[WIP] Implement Deployment of AWS CloudFormation templates

mmunozba:


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationScriptCreator.java
```diff
@@ -0,0 +1,55 @@
...
+import org.opentosca.toscana.plugins.scripts.EnvironmentCheck;
+
+/**
+ * Class for creating scripts needed for AWS CloudFormation deployment.
+ */
+public class CloudFormationScriptCreator {
+    private static final String CLI_COMMAND_CREATESTACK = "aws cloudformation deploy ";
+    private static final String CLI_PARAM_STACKNAME = "--stack-name ";
+    private static final String CLI_PARAM_TEMPLATEFILE = "--template-file ";
+    //TODO check if PARAMOVERRIDES are needed and get said parameters
```
mmunozba: I removed this from the command for now, since `--parameter-overrides` without any paramters won't start the depoyment.
I'm guessing we'll have to get this from the parameters in cfnModule?

mmunozba:
Since we decided to not use the AWS CLI in [ADR-0013](https://github.com/StuPro-TOSCAna/TOSCAna/blob/master/docs/adr/0013-aws-file-upload.md) this PR is obsolete.

mmunozba:
Since we decided to use the AWS CLI again to manage the file uploads, this PR can be recycled.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationScriptCreator.java
mmunozba: I added an issue for this: #413

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    /**
+     * Creates an S3Bucket with the given name.
+     * Wraps resources/cloudformation.scripts/create-bucket.sh
+     */
+    private String createBucket() {
+        // TODO get region, currently defaults to us-west-2
+        return "createBucket \"" + cfnModule.getBucketName() + "\"" + " \"us-west-2\"";
```
mklopp: You can do `cfnModule.getAWSRegion()` now

server/pom.xml
```diff
@@ -46,7 +46,7 @@
         <dependency>
             <groupId>com.github.StuPro-TOSCAna</groupId>
             <artifactId>cloudformation-builder</artifactId>
-            <version>72c1e0c2e80808d69b4aeb185a7533ebfa2ac3da</version>
+            <version>5b856331b4d8718e83bc715be4a827adbc1db61c</version>
```
mklopp: You could try this with the newest hash 1294a9af6c8dff82d7cca757ecbbdb2a004643c1

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /**
+     * Puts the given file with the given key on the S3Bucket with  the given name.
+     * Wraps resources/cloudformation.scripts/upload-file.sh
+     */
+    private String uploadFile(String objectKey, String filePath) {
+        return "uploadFile \"" + cfnModule.getBucketName() + "\" \"" + objectKey + "\" \"" + filePath + "\"";
```
mklopp: Don't know whether you actually need the `"` here

mmunozba: Going to check whether this is needed

mmunozba: After thinking a bit about this, this is just the normal passing of String parameters in bash.
The `""` are used for Strings that have spaces in them. Therefore you don't need them for the bucketName or any other parameters that do not have spaces in them. 
For objectKey and filePath: I believe that these could possibly contain spaces in their paths, so I'll be keeping the `""` around them.

mmunozba: After thinking a bit about this, I think this is just the normal passing of String parameters.
The "" are used for Strings that have spaces in them. Therefore you don't need them for the bucketName or any other parameters that do not have spaces in them.
For objectKey and filePath: I believe that these could possibly contain spaces in their paths, so I'll be keeping the "" around them.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -0,0 +1,173 @@
...
+
+import org.apache.commons.io.Charsets;
+import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+
+/**
+ * Class for building scripts and copying files needed for deployment of cloudformation templates.
+ */
+public class CloudFormationFileCreator {
+    static final String CLI_COMMAND_CREATESTACK = "aws cloudformation deploy ";
```
JensMueller2709: maybe you want to add a modifier.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -0,0 +1,173 @@
...
+    static final String CLI_PARAM_STACKNAME = "--stack-name ";
+    static final String CLI_PARAM_TEMPLATEFILE = "--template-file ";
+    //TODO check if PARAMOVERRIDES are needed and get said parameters
+    // private static final String CLI_PARAM_PARAMOVERRIDES = "--parameter-overrides ";
```
JensMueller2709: do you still need this comment?

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -0,0 +1,173 @@
...
+            logger.debug("Files to be copied found.");
+            logger.debug("Copying files to the target artifact.");
+            filesToBeUploaded.forEach((filePath) -> {
+                String targetPath = CloudFormationModule.FILEPATH_TARGET + filePath;
+                try {
+                    cfnModule.getFileAccess().copy(filePath, targetPath);
+                } catch (IOException e) {
+                    logger.error("Copying of files to the target artifact failed.");
+                    logger.error("See the stack trace for more info.");
+                    e.printStackTrace();
```
JensMueller2709: you could add the error message to the log before.
In general, maybe you could consider to use "one-line-logs".

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationLifecycle.java
```diff
@@ -71,13 +72,27 @@ public void transform() {
                     node.accept(cfnNodeVisitor);
                 }
             }
-
+            logger.info("Creating CloudFormation template.");
             fileAccess.access("output/template.yaml").appendln(cfnModule.toString()).close();
```
JensMueller2709: you could use the static OUTPUT variable of the AbstractLifecycle

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationModule.java
```diff
@@ -50,48 +55,68 @@
     private AWSCredentials awsCredentials;
 
     private Object keyNameVar;
-
     private Map<String, CFNInit> cfnInitMap;
-
+    private ArrayList<String> filesToBeUploaded;
```
JensMueller2709: Consider using just `List<String>`

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationModule.java
```diff
@@ -50,48 +55,68 @@
...
-     @param resource String id of a resource
+     * Get the CFNInit which belongs to a specific resource
+     *
+     * @param resource String id of a resource
      */
     public CFNInit getCFNInit(String resource) {
         return this.cfnInitMap.get(resource);
     }
 
+    public ArrayList<String> getFilesToBeUploaded() {
```
JensMueller2709: Also consider using `List<String>`

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
```diff
@@ -229,67 +228,74 @@ private void handleOperation(Operation operation, String serverName, String conf
...
-
-                // Add file to install
-                cfnModule.getCFNInit(serverName)
-                    .getOrAddConfig(CONFIG_SETS, config)
-                    .putFile(cfnFile);
-            } catch (IOException e) {
-                logger.error("Problem with reading file " + dependency);
-                e.printStackTrace();
-            }
+            String cfnFileMode = "000644"; //TODO Check what mode is needed (only read?)
```
JensMueller2709: Always fix values? Then you could make them `final static`

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
mmunozba: This was done to allow access for the test classes (no modifier makes them package-private).
I'm not sure if this is a good or bad practice though, so I don't know if it's better the add separate Strings to the test classes or to use the same for both like I did it here.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
mmunozba: Yes, this will be used in #413.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
mmunozba: Yes, for now, they are always fixed values, I changed them to final static accordingly.

In the future, this might be changed, see the TODOs and #406.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
mmunozba: It's probably best to just throw TransformationFailureExceptions here since these files are required in order for the transformation to be complete.

PR #377, opened 2017-12-19
Find transformation algorithm similarities between the different plugins

nfode:
To ensure that we do not have tons of duplicated code it is necessary to find a general algorithm that reduces the logic implemented in every plugin.

Finding a general algorithm there are two things every plugin has to do.

**1. Step:**
Go to the [example-models](https://github.com/StuPro-TOSCAna/TOSCAna/tree/feature/mdl/pseudo-code/docs/dev/effective-model/example-models) and copy the models [here](https://github.com/StuPro-TOSCAna/TOSCAna/tree/feature/mdl/pseudo-code/docs/dev/effective-model/general-node-sorter/pseudo-code) in the sub folder according to your plugin. Then enumerate all nodes in the models by the order you would transform them, if there is a step in which you "remove" a node you can also mark this in the model.

**2. Step:**
Write pseudo code that can transform all these models to your plugins target artifact. You can add the pseudo code in the document you can find in [here](https://github.com/StuPro-TOSCAna/TOSCAna/tree/feature/mdl/pseudo-code/docs/dev/effective-model/general-node-sorter/pseudo-code) in the sub folder according to your plugin. The pseudo code only has to be understandable so do not be overly artistic. Pseudo code how you exactly transform to the target artifact is not necessary, important is the pseudo code how you work with the graph.

If all plugins have finished the above steps I am going to create a general algorithm out of the results if possible.


c-mueller:
I honestly don't like the idea of considering such edge case topologies yet. I mean seriously. Most of the plugins don't support anything, except the things needed for the LAMP stack. We should first of all be able to support simpler topologies and other ndoe Types. Before even considering such models. (Some aspects of the topologies are important even now but this seems like overengineering) 

I also do not like the idea of pseudo code here it seriously is just a waste of time. The better idea is to let the plugins implement the mechanism imagined by them and then consider abstracting the common parts out. 

Apart from that. Model 1 cannot be deployed / transformed due the cyclic dependency. This represents a classical deadlock situation. 
This at least is the case if you look at it without knowing when they connect.
I mean this would in theory mean that WA1 Has to get started after WA2 has launched, but that probably is not possible, because WA2 tries to connect to WA1 that has not yet launched!

Code that detects things like Dependency Cycles can certanly get abstracted out, even at the current point though

nfode:
The cyclic dependency in the first one was with intention...

c-mueller:
> And if you read the spec a ConnectsTo is no dependency.

Really? I think we should ask the TOSCA Spec about that:


Taken from: [TOSCA-Simple-Profile-YAML-v1.1](http://docs.oasis-open.org/tosca/TOSCA-Simple-Profile-YAML/v1.1/csprd01/TOSCA-Simple-Profile-YAML-v1.1-csprd01.pdf) Page 200+

I Mean they are not equal. Especialy due tue the Inheritance hierarchy but we should probably consider them as "Semantically equivalent" therefore the problem stated above clearly is a Cylcic Dependency.

In this case the behaviour described in the Spec also represents the intuitive behaviour that i described above!

nfode:
If you have two nodes lets call them source and target, if there is a DependsOn from source to target target has to exist before source is created. If there is a ConnectsTo between them the direction does not matter but both nodes have to exist before the ConnectsTo relation is created l. So DependsOn and connects to are definitly not the same!

c-mueller:
As we decided to icebox this topic, i will close this PR

PR #379, opened 2017-12-19
Input retrieval from tosca template

hnicke:
Implement and test correct serving of input properties defined in TOSCA templates via the rest api.


server/src/integration/java/org/opentosca/toscana/core/parse/converter/ModelInstanceConverterIT.java
```diff
@@ -13,19 +13,25 @@
...
         assertNotNull(model);
     }
-    
+
+    // TODO WIP
+//    @Test
+//    public void lampInputConverter() throws Exception {
+//        EffectiveModel model = parser.parse(TestCsars.VALID_LAMP_INPUT_TEMPLATE);
+//        assertNotNull(model);
+//    }
```
hnicke: Won't work yet (#380)

c-mueller: is this still a thing even with the "Big Refactor"?

hnicke: Construction of the EffectiveModel after the big refactor looks like this:
```java
EffectiModel model = new EffectiveModel(TestCsars.VALID_LAMP_INPUT_TEMPLATE);
````
After the big refactor branch is applied, this will be uncommented again.

File: server/src/test/java/org/opentosca/toscana/core/BaseSpringTest.java [hunk 2]
```diff
@@ -50,12 +50,12 @@ private void initTmpdir() {
     public final void startupPrepareDisk() throws IOException {
         FileUtils.deleteDirectory(tmpdir);
         tmpdir.mkdirs();
-        logger.info("Cleaned up disk");
+        logger.info("Clean up disk");
```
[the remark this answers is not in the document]
hnicke: I revert the change.

PR #385, opened 2018-01-02
Service credential injection

JensMueller2709:


server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLifecycle.java
```diff
@@ -31,30 +40,43 @@ public boolean checkModel() {
...
+                provider.setOfferedService(cloudFoundryConnection.getServices());
+            }
+        }
+    }
+
+    private boolean isNotNull(String... elements) {
+        Boolean isOk = true;
+        for (String el : elements) {
+            if (el == null) {
+                isOk = false;
```
c-mueller: you dont need to iterate further if one value is null. The result of this method is false at that point. Therefore you can replace this by `return false;`

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLifecycle.java
```diff
@@ -31,30 +40,43 @@ public boolean checkModel() {
...
+            }
         }
+        return isOk;
```
c-mueller: When doing the thing described above. this can be replaced by `return true`

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLifecycle.java
```diff
@@ -64,8 +86,13 @@ public void transform() {
...
         try {
-            CloudFoundryFileCreator fileCreator = new CloudFoundryFileCreator(context.getPluginFileAccess(), myApp);
+            CloudFoundryFileCreator fileCreator = new CloudFoundryFileCreator(fileAccess, myApp);
             fileCreator.createFiles();
+            if (cloudFoundryConnection != null) {
+                CloudFoundryInjectionHandler injectionHandler = new CloudFoundryInjectionHandler(fileAccess, myApp);
+                injectionHandler.injectServiceCredentials();
+                fileCreator.updateManifest();
+            }
         } catch (IOException e) {
```
c-mueller: Does this exception and the next need different treatment? It not, please combine them using `catch (IOException | JSONException e)` (the syntax might differ)

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/application/CloudFoundryApplication.java
```diff
@@ -54,6 +66,14 @@ public void addEnvironmentVariables(String environmentVariableName) {
         return services;
     }
 
+    public ArrayList<CloudFoundryService> getServicesMatchedToProvider() {
```
c-mueller: Replace this with List

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/application/CloudFoundryApplication.java
```diff
@@ -54,6 +66,14 @@ public void addEnvironmentVariables(String environmentVariableName) {
...
+        return servicesMatchedToProvider;
```
c-mueller: Do you need write access on this list? if not: Consider making it read only using  [`Collections.unmodifiableList()`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#unmodifiableList-java.util.List-)

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/client/CloudFoundryInjectionHandler.java
```diff
@@ -0,0 +1,100 @@
...
+            if (cloudFoundryConnection != null) {
+                Path pathToApplication = Paths.get(fileAccess.getAbsolutePath(app.getPathToApplication()));
+                succeed = cloudFoundryConnection.pushApplication(pathToApplication,
+                    app.getName(), app.getServicesMatchedToProvider());
+            }
+        } catch (InterruptedException e) {
+            logger.error("Something went wrong while pushing the application");
+            logger.error(e.getMessage());
+        } catch (FileNotFoundException e) {
+            logger.error("Application could not be found while pushing the application");
```
c-mueller: I dont like the sound of this error message. FileNotFoundExeption tells me that the file is probably missing on the local filesystem. Please consider adding such information using a short sentence

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/client/CloudFoundryInjectionHandler.java
```diff
@@ -0,0 +1,100 @@
...
+    public void injectServiceCredentials() {
+        deploy();
+        getServiceCredentials();
+    }
+
+    /**
+     deploys the application, creates the services and bind the application to the service
+     the application is not running after that
+     */
+    public Boolean deploy() {
```
c-mueller: Do you need a Boxed Boolean? If not please use primitives (`boolean`) instead

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/client/CloudFoundryInjectionHandler.java
```diff
@@ -0,0 +1,100 @@
...
+        Boolean succeed = false;
+        try {
+            if (cloudFoundryConnection != null) {
+                Path pathToApplication = Paths.get(fileAccess.getAbsolutePath(app.getPathToApplication()));
+                succeed = cloudFoundryConnection.pushApplication(pathToApplication,
+                    app.getName(), app.getServicesMatchedToProvider());
+            }
+        } catch (InterruptedException e) {
+            logger.error("Something went wrong while pushing the application");
+            logger.error(e.getMessage());
```
c-mueller: Please print the stacktrace of the exception. Just printing the message is like saying "The internet does not work" without giving further information, this makes debugging and understanding Bugs a horrible experience.

A simple solution is replacing the two lines with the following:
```Java
logger.error("Something went wrong while pushing the application", e);
```

server/src/test/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryServiceTest.java
```diff
@@ -52,40 +72,116 @@ public void setUp() {
...
+    private void setUpMyApp() throws IOException, JSONException {
+        CloudFoundryNodeVisitor visitor = new CloudFoundryNodeVisitor(myApp);
+        EffectiveModel lamp = TestEffectiveModels.getLampModel();
+        ArrayList<String> paths = new ArrayList<>();
+        String resourcesPath = "src/test/resources/";
+        File sourceDir = new File(resourcesPath + "csars/yaml/valid/lamp-input/");
         targetDir = new File(tmpdir, "targetDir");
         sourceDir.mkdir();
         targetDir.mkdir();
+        when(log.getLogger(any(Class.class))).thenReturn(LoggerFactory.getLogger("Test logger"));
```
c-mueller: There is a util method that allows mocking of the Log Class:
https://github.com/StuPro-TOSCAna/TOSCAna/blob/master/server/src/test/java/org/opentosca/toscana/core/testutils/TestUtils.java

server/src/integration/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLampIT.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    protected void onFailure(File outputDir, Exception e) throws Exception {
+        fail();
+    }
+
+    @Override
+    protected PropertyInstance getProperties() throws Exception {
+        String envUser = System.getenv("TEST_CF_USER");
```
c-mueller: Maybe also consider putting the env keys in Constants

PR #386, opened 2018-01-04
Fix common LAMP Model

c-mueller:


server/src/test/resources/csars/yaml/valid/lamp-input/my_app/create_myphpapp.sh
```diff
@@ -1,7 +1,4 @@
 #!/bin/bash
-# install php on a linux machine with php-mysql
```
mklopp: This is the installation of php not apache and it is mandatory because no where else is stated that php has to be installed on the websever

c-mueller: Well i failed there (kind of). In order to simplify the transformation on kubernetes (the installation within the webapp create script causes issues) id suggest putting this in the Create Apache script.

c-mueller: It actually was a duplicate. just look at the install apache script. 
image

However i dont like the idea that apache is only installed implicitly (if it even is)

mklopp: You are right that code is also in `install_apache.sh` but this file is never referenced in the template.yaml .
I actually like the idea that apache is installed implicitly. The WebServer.Apache node means that there should be a webserver running with apache installed.

mklopp:
I added the install php script as the configure lifecycle for the apache node. Maybe you can give your opinion on that.

c-mueller:
This probably is a good compromise

PR #390, opened 2018-01-04
Fix aws to support common lamp model

mklopp:


mklopp:
I actually already looked at that. It's a cloudfoundry test from @JensMueller2709 that is failing to to the changes to the common lamp model

mmunozba:
Is there a way to fix this? Can we maybe move some of the changes from this PR to #386?

c-mueller:
Can we merge this?

mklopp:
with the new changes to common lamp model this is wip again

c-mueller:
Considering this done. I am therefore merging it.

PR #391, opened 2018-01-05
Generalize away from the lamp app

mklopp:


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+        .build();
+
+    /**
+     This method requests the AWS server for ImageIds with filters which are filled based on
+     the values of the OsCapability. The image with the latest creation date is picked and its imageId returned.
+
+     @param osCapability The OsCapability to map.
+     @return A String that contains a valid ImageId that can be added to the properties of an ec2.
+     */
+    public String mapOsCapabilityToImageId(OsCapability osCapability) {
```
JensMueller2709: complex method, maybe you could extract some logic to additional methods

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+                    .toString() + "*"));
+            }
+        }
+        if (osCapability.getVersion().isPresent()) {
+            describeImagesRequest.withFilters(new Filter("name").withValues("*" + osCapability.getVersion().get() +
+                "*"));
+        }
+        if (osCapability.getArchitecture().isPresent()) {
+            if (osCapability.getArchitecture().get().equals(OsCapability.Architecture.x86_64)) {
+                describeImagesRequest.withFilters(new Filter("architecture").withValues("x86_64"));
```
JensMueller2709: you could use constant fields for "x86_64". Same for "i386"

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+                Image latest = creationDateMap.get(Collections.max(creationDateMap.keySet()));
+                logger.debug("Latest image received: " + latest.toString());
+                imageId = latest.getImageId();
+            } else {
+                logger.warn("No images received defaulting to old ubuntu 16.04 image");
+                imageId = "ami-0def3275";
+                //TODO maybe not defaulting but throwing a transformation failed exception?
+            }
+        } catch (SdkClientException se) {
+            logger.error("Cannot connect to AWS to request image Ids, defaulting to old ubuntu 16.04 image");
```
JensMueller2709: Is there a problem if the user want a Windows system and you set the default to a ubuntu version?

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+            .sorted()
+            .collect(Collectors.toList());
+        // scale numCpus and memSize upwards if they are not represented in the lists
+        try {
+            logger.debug("Check numCpus: " + numCpus);
+            numCpus = checkValue(numCpus, allNumCpus);
+            logger.debug("Check memSize: " + memSize);
+            memSize = checkValue(memSize, allMemSizes);
+        } catch (IllegalArgumentException ie) {
+            String errorMessage = "Values numCpus: " + numCpus + " and/or memSize: " + memSize + " are to big. No " +
```
JensMueller2709: you could use the String.format method

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+                "InstanceType found";
+            logger.error(errorMessage);
```
JensMueller2709: I think it is sufficient to add the errorMessage to the `TransformationFailureException`

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+     @return A String that contains a valid ImageId that can be added to the properties of an ec2.
+     */
+    public String mapOsCapabilityToImageId(OsCapability osCapability) {
+        AmazonEC2 ec2 = AmazonEC2ClientBuilder.standard()
+            .withRegion(Regions.US_WEST_2) //TODO get this from user
+            .build();
+        //need to set these, owner are self and amazon
+        DescribeImagesRequest describeImagesRequest = new DescribeImagesRequest()
+            .withFilters(
+                new Filter("virtualization-type").withValues("hvm"),
```
mmunozba: Are these always the same or do we have to set them dynamically in the future?
Same for most of the other hard-coded Strings in the Filters.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
mmunozba: Agreed, for example mapping the ImageId could probably be extracted. (see below)
Also filling the describeImageRequest could probably be a separate method.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+            if (osCapability.getArchitecture().get().equals(OsCapability.Architecture.x86_64)) {
+                describeImagesRequest.withFilters(new Filter("architecture").withValues("x86_64"));
+            } else if (osCapability.getArchitecture().get().equals(OsCapability.Architecture.x86_32)) {
+                describeImagesRequest.withFilters(new Filter("architecture").withValues("i386"));
+            }
+        } else {
+            //defaulting to 64 bit architecture
+            describeImagesRequest.withFilters(new Filter("architecture").withValues("x86_64"));
+        }
+        String imageId;
```
mmunozba: You could probably extract this part until the return statement and return something like mapImageId(...) instead.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+
+     @param computeCapability The ComputeCapability to map.
+     @param distinction       A distinction string. Can either be "EC2" or "RDS".
+     @return A valid InstanceType / InstanceClass string.
+     @throws TransformationFailureException Gets thrown if the values numCpus and memSize are too big and there is no
+     valid InstanceType.
+     */
+    public String mapComputeCapabilityToInstanceType(ComputeCapability computeCapability, String distinction) throws
+        TransformationFailureException {
+        //TODO what to do with disk size?
```
mmunozba: Is this maybe possible with attached volumes? Can you change the root volume of the EC2 instance?
Added an issue for this: #423

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
mmunozba: "to big" -> "too big"

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+        return instanceType;
+    }
+
+    /**
+     The value is taken from the Capability but turned into GB. The minimum is 20 GB the maximum 6144 GB
+
+     @param computeCapability The ComputeCapability to map.
+     @return An integer representing the diskSize that should be taken.
+     */
+    public Integer mapComputeCapabilityToDiskSize(ComputeCapability computeCapability) {
```
mmunozba: If I understand this method properly, this converts the disk size to GB and caps it if it exceeds the min or max value right?
If so, I think the method name could better represent the functionality of the method.

I commented earlier about the TODO regarding disk_size. From what I've seen, this method currently only gets used for setting the AllocatedStorage of SQL nodes hosted on a Compute node. Since the min/max sizes are probably different for the EC2 instances, we should probably make it clear that this method is only used for the SQL nodes.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+        }
+        if (diskSize < minSize) {
+            logger.debug("Disk size: " + minSize);
+            return minSize;
+        }
+        logger.debug("Disk size: " + diskSize);
+        return diskSize;
+    }
+
+    private Integer checkValue(Integer value, List<Integer> checker) throws IllegalArgumentException {
```
mmunozba: You should probably add comments or change the name for the following two methods, "checkValue" and "findCombination" don't tell me anything about their functionality.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+            .filter(u -> u.getNumCpus().equals(numCpus) && u.getMemSize().equals(memSize))
+            .findAny();
+        if (instanceType.isPresent()) {
+            return instanceType.get().getType();
+        } else {
+            return "";
+        }
+    }
+
+    private class InstanceType {
```
mmunozba: Isn't this something that should be added to the cloudformation-builder?

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
```diff
@@ -67,29 +67,14 @@ public void visit(Compute node) {
...
-            if (computeOs.getType().get().equals(OsCapability.Type.LINUX) &&
-                computeOs.getDistribution().get().equals(OsCapability.Distribution.UBUNTU) &&
-                computeOs.getVersion().get().equals("16.04")) {
-                imageId = "ami-0def3275";
-            } else {
-                throw new UnsupportedTypeException("Only Linux, Ubuntu 16.04 supported.");
-            }
+            String imageId = capabilityMapper.mapOsCapabilityToImageId(computeOs);
             //check what host should be taken
             // we only support t2.micro atm since its free for student accounts
```
mmunozba: This isn't true anymore, right?

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
```diff
@@ -182,6 +174,10 @@ public void visit(Apache node) {
             Operation configure = node.getStandardLifecycle().getConfigure().get();
             handleOperation(configure, serverName, CONFIG_CONFIGURE);
         }
+        //we add restart apache2 command to the configscript
```
mmunozba: Is this always needed?

server/pom.xml
```diff
@@ -169,19 +190,12 @@
             <version>1.0-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
-        <!--Apache Http Client (Might not be needed anymore.-->
-        <!--Did not remove it because its just a test dependency-->
+        <!--Apache Http Client -->
+        <!-- needed for AWS Java SDK-->
```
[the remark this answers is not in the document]
mmunozba: Apparently, this was not used besides for initial testing so it should not affect any other functionality.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
```diff
@@ -0,0 +1,310 @@
...
+import com.amazonaws.services.ec2.model.Image;
+import com.google.common.collect.ImmutableList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CapabilityMapper {
+
+    private final static Logger logger = LoggerFactory.getLogger(CapabilityMapper.class);
+
+    private final ImmutableList<InstanceType> EC2_INSTANCE_TYPES = ImmutableList.<InstanceType>builder()
```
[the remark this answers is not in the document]
mklopp: I only found [this](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/ec2/model/InstanceType.html) but these have no correlation to their numCpus and memSize

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
mklopp: No its not always needed, but there is no problem in restarting it just before the deployment is finished.
If there are modules installed that apache2 correlates with (like php) we need a restart.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/mapper/CapabilityMapper.java
mklopp: Maybe, but its not concerning the template directly so i think its fine here

PR #392, opened 2018-01-06
[WIP] Implement Upload of files to AWS S3Bucket

mmunozba:
From #392.


mmunozba:
Since we are adding the File Uploads through scripts/the AWS CLI, this PR is no longer needed since the implementation was done with the AWS Java SDK.

PR #398, opened 2018-01-09
Replace restricted strings in app name and add random-route

### LINKED ISSUE #394: Application name must not contain "_"

JensMueller2709:
resolves #394


server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/FileCreator.java
```diff
@@ -222,6 +233,6 @@ private void addProviderServiceOfferings(BashScript deployScript) throws IOExcep
             }
             deployScript.append(String.format("# %-20s %-40s %-50s ", service.getLabel(), plans, service.getDescription()));
         }
-        deployScript.append("\n* These service plans have an associated cost. Creating a service instance will incur this cost.");
+        deployScript.append("\n# * These service plans have an associated cost. Creating a service instance will incur this cost.");
```
mmunozba: Shouldn't this just be `# These service...` instead of `\n# * These service...`

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/application/Application.java
```diff
@@ -14,7 +14,10 @@
  */
 public class Application {
 
+    //private static int counter;
```
mmunozba: Why is this here? Same with the other counter-related additions

JensMueller2709: already deleted

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/FileCreator.java
JensMueller2709: No, I need an additional new line and the `*` is also needed as annotation.

PR #399, opened 2018-01-09
Rework property setting mechanism

### LINKED ISSUE #127: Replace Property Setting mechanism
To Set the properties currently you have to send a response like this to the server:
```Json
{
	"properties": {
		"text_property":"Hallo Welt",
		"name_property": "hallo"
	}
}
```

But instead we want to send the Original JSON (Returned by the GET Request):
```Json
{
    "properties": [
        {
            "key": "text_property",
            "type": "text",
            "required": true,
            "description": ""
        },
        {
            "key": "name_property",
            "type": "name",
            "required": true,
            "description": ""
        }
    ],
    "_links": {
        "self": {
            "href": "http://localhost:8080/csars/mongo-db/transformations/p-a/properties"
        }
    }
```

with the values attached to each inner property element:
```Json
{
    "properties": [
        {
            "key": "text_property",
            "type": "text",
            "required": true,
            "description": "",
            "value": "Hallo"
        },
        {
            "key": "name_property",
            "type": "name",
            "required": true,
            "description": "",
            "value": "myname"
        }
    ],
    "_links": {
        "self": {
            "href": "http://localhost:8080/csars/mongo-db/transformations/p-a/properties"
        }
    }
```

nfode:
closes #127


retrofit-wrapper/src/test/java/org/opentosca/toscana/retrofit/api/TransformationInputTest.java
```diff
@@ -0,0 +1,5 @@
+package org.opentosca.toscana.retrofit.api;
+
+public class TransformationInputTest extends BaseTOSCAnaAPITest {
```
c-mueller: Will this be impelemented?

PR #401, opened 2018-01-10
Change db_user to a environment variable in common Lamp

### LINKED ISSUE #396: Environment variable "database_user" in Lamp-App
In the [LampApp implementation](https://github.com/StuPro-TOSCAna/TOSCAna/blob/master/server/src/test/java/org/opentosca/toscana/plugins/testdata/LampApp.java) there is a input for the "database_user" (Line 139):
` appInputs.add(new OperationVariable("database_user", database.getUser().get()));`
In the file [mysql-credentials.php](https://github.com/StuPro-TOSCAna/TOSCAna/blob/master/server/src/test/resources/csars/yaml/valid/lamp-input/my_app/mysql-credentials.php) there is a fix input for the user:
`$db_user = "root";`

The Cloud Foundry plugin can´t set userprovided database credentials. Instead it provide the database credentials as environment variables. 
So my suggestion is to set the `$db_user` also as environment variable. Therefore we need a additional line in [configure_myphpapp.sh](https://github.com/StuPro-TOSCAna/TOSCAna/blob/master/server/src/test/resources/csars/yaml/valid/lamp-input/my_app/configure_myphpapp.sh):
`sed -i "s:DATABASE_USER:${database_user}:g" $CREDENTIALS`

@c-mueller unfortunately I saw this issue as of now.

If it is ok for every plugin I will change the files

JensMueller2709:
resolves #396


server/src/test/resources/csars/yaml/valid/lamp-input/template.yaml
```diff
@@ -54,6 +55,7 @@ topology_template:
       type: tosca.nodes.Database.MySQL
       properties:
         name: { get_input: database_name }
+        user: root
```
c-mueller: If we consider doing it dynamically then, this should also be a input

PR #404, opened 2018-01-11
Implement basic Pod grouping

c-mueller:
This PR Implements the Grouping of pods (by compute node). i.e. All containers that originally belonged to the same compute node will land in the same pod.

This feature will need updating once we migrate to DockerApplications instead of nodestacks, but for now it works.

This also has been the last step to allow the transformation of the Common LAMP model.
This has been verified! Deployment works!


server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/KubernetesLampIT.java
```diff
@@ -40,7 +40,7 @@ protected void checkAssumptions() {
     @Override
     protected void onSuccess(File outputDir) throws Exception {
         //Comment out the sleep command to allow the interruption of the test before the files get deleted
-        Thread.sleep(10000);
+        Thread.sleep(10000+0);
```
nfode: Plz delete this.

PR #407, opened 2018-01-12
Implement Multi-Compute Support for Kubernetes Plugin

c-mueller:
The goal here is to Support the LAMP Model with multiple  Compute nodes. That means that the database and the Webserver will be placed in different pods.


c-mueller:
Due to this actually beeing completely operational. I thunk we hould merge this. To allow the next stpest to be done in a new branch

Issue #395, opened 2018-01-08
Execute shell scripts in warden container


JensMueller2709:
To provide configure scripts like the "configure_myphpapp.sh" of the Lamp-App, it is needed to execute shell scripts in the cloud foundry instance.

Issue #375, opened 2017-12-19
Provide SQL file to MySQL Service to create tables


orferch:
Investigate how to provide SQL files to the Service, to create the wanted tables after the Service is started.

JensMueller2709:
This with this [project](https://github.com/andreasf/cf-mysql-plugin) we could connect to a mysql database and could execute a sql-File. How we already discussed, we try to use it.

orferch:
Currently working Solution:

on Server:
- App is connected to the MySQL Service

on Local Machine:
- cloud-foundry cli is installed
- install cf-mysql-plugin with `cf install-plugin -r "CF-Community" mysql-plugin`
- install mysql-server

move into the folder with the *.sql file and connect to the created MySQL Service:
- show available Services: `cf mysql`
- connect to the Database: `cf mysql mydbname`
- provide the .sql file with `source filename.sql`

MySQL runs now with the provided Information of the .sql


We will look more into running Scripts to automate this whole process.

PR #421, opened 2018-01-17
Extract deployment from transformation process

### LINKED ISSUE #408: Extract deployment logic of the transformation
To push files to the CF instance while the transformation process is not a good way because we hide this from the user and create e.g. services without his knowing.
To avoid this we have to extract this logic into the deployment-scripts. Consequently, we have to do a bigger rework to the java-implementation. 
- [x] no deploying while transformation
- [x] provide scripts for readingCredentials and so on
- [x] add code to bash script. 

EDIT: "Implement logic to transformation" I will create a new PR for this because it might be bigger changes.

A first approach of the used scripts you could find [here](https://hackmd.io/s/B14EwBL4f#)

JensMueller2709:
resolves #408 , #375 , #395


server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/application/deployment/Deployment.java
```diff
@@ -0,0 +1,73 @@
...
+    private final String PYTHON_EXECUTE_FILE = "/cloudFoundry/deployment_scripts/executeCommand.py";
+    private final String PYTHON_EXECUTE_FILE_TARGET = "scripts/executeCommand.py";
+    private final String PYTHON_READ_CREDENTIALS = "/cloudFoundry/deployment_scripts/readCredentials.py";
+    private final String PYTHON_REPLACE_STRINGS = "/cloudFoundry/deployment_scripts/replace.py";
+
+    public Deployment(BashScript deploymentScript, Application application, PluginFileAccess fileAccess) throws IOException {
+        this.deploymentScript = deploymentScript;
+        this.application = application;
+        this.fileAccess = fileAccess;
+        clazz = Deployment.class;
```
orferch: rename clazz

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/client/InjectionHandler.java
```diff
@@ -48,7 +50,8 @@ public boolean deploy() {
         boolean succeed = false;
         try {
             if (connection != null) {
-                Path pathToApplication = Paths.get(fileAccess.getAbsolutePath(app.getPathToApplication()));
+                Path pathToApplication = Paths.get(fileAccess.getAbsolutePath(APPLICATION_FOLDER +
```
orferch: Stringbuilder would be better here

JensMueller2709:
this PR is still on WIP, so I am working on it. But of course I will consider your requested changes

docs/dev/plugins/cloudFoundry/Script-Overview.md
```diff
@@ -0,0 +1,41 @@
...
+## Order
+1. `cf create service $servicename $serviceplan $serviceInstanceName`
+2. `cf push $appName -f $pathToManifest`
+3. `python readCredentials.py $appName $serviceName $serviceType`
+4. `python replace.py $file $findString $replaceString`
+5. `python executeScript.py $pathInWardenContainter $appName`
+6. `python configureMysql.py $sqlFile` or 12Factor: `python executeScript.py $appName $dbInitScript`
+
+## Summary
+the AWS Cloudfourmation CLI needs python anyways so we could use it too. We think for "bigger" script implementation python is the best way to do it.
```
mklopp: Cloudformation

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/application/deployment/Deployment.java
```diff
@@ -0,0 +1,133 @@
...
+
+public class Deployment {
+
+    private BashScript deploymentScript;
+    private Application application;
+    private PluginFileAccess fileAccess;
+    private Class deploymentClass;
+    private boolean pythonIsChecked;
+
+    private final String PYTHON_SCRIPTS_TARGET = SCRIPTS_DIR_PATH;
```
mklopp: These should be static?

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/application/deployment/Deployment.java
```diff
@@ -0,0 +1,133 @@
...
+    private boolean isAlreadyCopied(String fileName) {
+        try {
+            fileAccess.getAbsolutePath(PYTHON_SCRIPTS_TARGET + fileName);
+        } catch (FileNotFoundException e) {
+            return false;
+        }
+        return true;
+    }
+
+    private void checkPython() throws IOException {
```
mklopp: You do this `checkPyhton()` after every call to `copyFile()` so you can just include it that method

server/src/main/resources/cloudFoundry/deployment_scripts/configureMysql.py
```diff
@@ -0,0 +1,47 @@
...
+import ast
+import sys
+
+# needs a special python mysql connector installation
+def main():
+    mysqlConfigFile = 'mysqlConfig.txt'
+    # should be a sql file
+    strConfigureFile = str(sys.argv[1])
+
+    if ".sql" not in strConfigureFile[len(strConfigureFile)-4:]:
```
mklopp: consider using: `".sql" == os.path.splitext(strConfigureFile)[-1].lower()` or `strConfigureFile.lower().endswith(".sql")`

server/src/main/resources/cloudFoundry/deployment_scripts/configureMysql.py
```diff
@@ -0,0 +1,47 @@
...
+        print("needs an .sql File for db init")
+        exit(1)
+
+    try:
+        with open(strConfigureFile, 'r') as dbinit_file:
+            dbinit_command = dbinit_file.read()
+    except IOError as err:
+        print("Failed to read dbinit File named %s. Where it is?" %(strConfigureFile))
```
mklopp: consider using `print("Failed to read dbinit File named {}. Where it is?".format(strConfigureFile))`

server/src/main/resources/cloudFoundry/deployment_scripts/configureMysql.py
```diff
@@ -0,0 +1,47 @@
+import mysql.connector
```
mklopp: how do you know this package is installed?

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/application/deployment/Deployment.java
JensMueller2709: Done

PR #430, opened 2018-01-22
Evaluate Relationships

mklopp:
Implement:
- [x] CheckModel
- [x] Prepare
- [x] VisitRelationships


server/src/integration/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationLampIT.java
```diff
@@ -31,7 +33,10 @@ protected void onSuccess(File outputDir) {
 
     @Override
     protected void onFailure(File outputDir, Exception e) {
-        fail();
+        //if the cause for transformation failure is amazonEC2Exception its fine because of the credentials 
+        if (!(e instanceof TransformationFailureException && e.getCause() instanceof AmazonEC2Exception)) {
```
c-mueller: Please consider using assumptions to prevnt a test from beeing executed if the credentials are not set.

you can use the check assumptions method in this test to do this.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationLifecycle.java
```diff
@@ -74,9 +86,16 @@ public void transform() {
...
+            logger.error("File access error");
+            throw new TransformationFailureException("Could not write with fileAccess", ie);
+        } catch (TransformationFailureException tfe) {
             logger.error("Transformation to CloudFormation unsuccessful. Please check the StackTrace for more Info.");
-            e.printStackTrace();
+            tfe.printStackTrace();
+            throw tfe;
+        } catch (Exception e) {
+            logger.error("Transformation to CloudFormation unsuccessful. Random Exception should not appear here.");
+            throw new TransformationFailureException("Random exception", e);
```
c-mueller: I dont like this wording. Unexpected error sonds much better.

Noting is really random here

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationLifecycle.java
```diff
@@ -74,9 +86,16 @@ public void transform() {
...
             fileAccess.access("output/template.yaml").appendln(cfnModule.toString()).close();
             logger.info("Transformation to CloudFormation successful.");
-        } catch (Exception e) {
+        } catch (IOException ie) {
+            logger.error("File access error");
+            throw new TransformationFailureException("Could not write with fileAccess", ie);
+        } catch (TransformationFailureException tfe) {
             logger.error("Transformation to CloudFormation unsuccessful. Please check the StackTrace for more Info.");
-            e.printStackTrace();
+            tfe.printStackTrace();
```
c-mueller: there is no need to print the stacktrace, because the ExecutionTask (the part of the core that handles the excution of transformation)

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
```diff
@@ -147,72 +148,93 @@ public void visit(MysqlDatabase node) {
...
+                handleOperation(configure, computeName, CONFIG_CONFIGURE);
+            }
+            //handle start
+            if (node.getStandardLifecycle().getStart().isPresent()) {
+                Operation start = node.getStandardLifecycle().getStart().get();
+                handleOperation(start, computeName, CONFIG_START);
+            }
+        } catch (Exception e) {
+            logger.error("Error while creating WebApplication");
+            throw new TransformationFailureException("Failed at WebApplication node", e);
```
c-mueller: Maybe condider adding the name of the node

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CheckModelNodeVisitor.java
```diff
@@ -0,0 +1,70 @@
...
+            OsCapability.Distribution distribution = osCapability.getDistribution().get();
+            if (!supportedDistributions.contains(distribution)) {
+                throw new UnsupportedTypeException("OS distribution " + distribution + " not supported.");
+            }
+        }
+    }
+
+    @Override
+    public void visit(MysqlDatabase node) {
+        logger.info("Check MysqlDatabase node {}.", node.getEntityName());
```
hnicke: I adopted the habit to always quote variables when logging.
```java
logger.info("Check MysqlDatabase node '{}'"),
```

This way you can detect null variables when reading the log:
```
INFO   Check MysqlDatabase node ''
```
But taste varies.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelNodeVisitor.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+        if (!node.getPort().isPresent()) {
+            logger.warn("Database port not set, setting to default");
+            node.setPort(DEFAULT_PORT);
+        }
+
+        // check if Mysql is the only node hosted on his compute node
+        Compute compute = getCompute(node);
+        if (topology.incomingEdgesOf(compute).size() == 1) {
+            // means our dbms is the only one hosted on this compute
```
hnicke: This is incorrect. The graph's edges are a Set of all possible Relationships. HostedOn is only one of them.
A quick fix could be:
```java
if ( topology.incomingEdgesOf(compute).stream().filter(p -> p instanceof HostedOn).collect(Collectors.toSet()).size() == 1)
```
Didn't try that code, though.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelRelationshipVisitor.java
```diff
@@ -0,0 +1,51 @@
...
+        RootNode target = topology.getEdgeTarget(relation);
+        if (source instanceof WebApplication && target instanceof MysqlDatabase) {
+            MysqlDatabase mysqlDatabase = (MysqlDatabase) target;
+            WebApplication webApplication = (WebApplication) source;
+            // if they are hosted on the same compute --> we can set the compute private address to
+            Compute computeMysqlDatabase = getCompute(mysqlDatabase);
+            Compute computeWebApplication = getCompute(webApplication);
+            if (computeMysqlDatabase.equals(computeWebApplication)) {
+                // means we can set the private address as reference the database endpoint
+                //TODO only set privateAddress or also publicAddress?
```
hnicke: Why not set the public address as well? Go for it.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelRelationshipVisitor.java
```diff
@@ -0,0 +1,51 @@
...
+                computeMysqlDatabase.setPrivateAddress(Fn.fnGetAtt(toAlphanumerical(mysqlDatabase.getEntityName()),
```
hnicke: Why is the private address set in both Visitor classes?

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelNodeVisitor.java
mklopp: Actually I'm pretty sure this is correct. Sure it is a Set of all possible relationships but I already know that the mysqlDatabase node I am currently visiting is hosted on this specific compute (or mysqlDatabase is hosted on mysqlDbms is hosted on this compute node).
That means if the compute node has exactly one relationship it is that hosted on relation.

hnicke: Imagine the tosca modeller has written a template that defines a timing dependency (DependsOn) from node 'otherNode' to 'server'.
Now, 
```java
if (topology.incomingEdgesOf(compute).size() == 1) {
```
..will evaluate to false, altough mysql dbms node is still the only node hosted on the server node.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelRelationshipVisitor.java
mklopp: I try to cover two possibilities here.
1. `PrepareModelNodeVisitor.java` sets the private address of the compute node to the database endpoint exactly when the database is the only node hosted on the compute. That means this compute node completely belongs to the database and represents it (by providing the endpoint)
2. `PrepareModelRelationshipVisitor.java` is used in the case of our current lamp-model where both the webapplication and the database are hosted on one compute. I only feel safe to set the private address of a compute node to reference the database if there is a connects to relationship between the two nodes sitting on top of the compute node.
Here I see the problem tackled in the previous meeting. What if someone references the webapplication as well by calling the private address of compute node it is hosted on (which is valid). He would get the database endpoint instead.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelNodeVisitor.java
mklopp: True

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelRelationshipVisitor.java
hnicke: I just had following alternative idea:
- when handling the `dbms` node;
  1. store the initial value of `compute#privateAddress`
  2. apply the AWS `Fn` fix (`compute.setPrivateAddress(Fn....)`)
  3. write all mysql dbms related configuration to aws template file
  4. restore value of `compute#privateAddress`
  
Could this work?

mklopp: I will look into this

mklopp: I don't think this will work. I actually need the privateAddress to be set to the database endpoint when I visit the webApplication node and not the dbms/database node.
I think the current solution is the best there is and will work for most cases.

mklopp:
The `TestCsars.VALID_LAMP_NO_INPUT_MULTI_COMPUTE_TEMPLATE` should work with this pr as well. The resulting template looks good but I can't test the deployment atm

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CheckModelNodeVisitor.java
```diff
@@ -0,0 +1,79 @@
...
+    public void visit(Apache node) {
+        logger.info("Check Apache node '{}'.", node.getEntityName());
+    }
+
+    @Override
+    public void visit(Compute node) {
+        logger.info("Check Compute node '{}'.", node.getEntityName());
+        List<OsCapability.Type> supportedTypes = new ArrayList<>();
+        supportedTypes.add(OsCapability.Type.LINUX);
+        //might grow but for now only linux
```
mmunozba: Added an issue for this: #462.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelNodeVisitor.java
```diff
@@ -0,0 +1,109 @@
...
+     @param cfnModule module to modify
+     */
+    public PrepareModelNodeVisitor(TransformationContext context, CloudFormationModule cfnModule) {
+        this.logger = context.getLogger(getClass());
+        this.topology = context.getModel().getTopology();
+        this.cfnModule = cfnModule;
+    }
+
+    @Override
+    public void visit(MysqlDatabase node) {
```
hnicke: Maybe refactor this, as its a quite big method.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
```diff
@@ -45,85 +54,118 @@
...
         this.cfnModule = cfnModule;
     }
 
+    /**
+     Get the Compute node this webApplication is ultimately hosted on
+
+     @param webApplication the webApplication to find the host for
+     @return the underlying Compute node
+     */
+    protected static Compute getCompute(WebApplication webApplication) {
```
mmunozba: I feel like `CloudFormationNodeVisitor` is getting pretty big. 
Maybe we should extract some of the utility methods like these `getCompute` methods? 
We might need these for other future visitors, so it might be a good idea to extract them regardless of the size of this class.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationLifecycle.java
```diff
@@ -10,59 +10,97 @@
...
     public boolean checkModel() {
-//        TODO implement model checks
+        logger.info("Check model for compatibility to CloudFormation");
+        Set<RootNode> nodes = model.getNodes();
+        Set<RootRelationship> relationships = model.getTopology().edgeSet();
+        try {
+            CheckModelNodeVisitor checkModelNodeVisitor = new CheckModelNodeVisitor(context);
+            logger.debug("Check nodes");
+            for (VisitableNode node : nodes) {
+                node.accept(checkModelNodeVisitor);
```
hnicke: In the Visitor classes you always log which node you are visiting when entering the method. Why don't you just write it here once for all methods?

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationLifecycle.java
```diff
@@ -10,59 +10,97 @@
...
+            }
+            CheckModelRelationshipVisitor checkModelRelationshipVisitor = new CheckModelRelationshipVisitor(context);
+            logger.debug("Check relationships");
+            for (VisitableRelationship relationship : relationships) {
+                relationship.accept(checkModelRelationshipVisitor);
```
hnicke: Same for relationships.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelRelationshipVisitor.java
```diff
@@ -0,0 +1,60 @@
...
+        logger.info("Prepare ConnectsTo relation '{}'.", relation.getEntityName());
+        RootNode source = topology.getEdgeSource(relation);
+        RootNode target = topology.getEdgeTarget(relation);
+        if (source instanceof WebApplication && target instanceof MysqlDatabase) {
+            MysqlDatabase mysqlDatabase = (MysqlDatabase) target;
+            WebApplication webApplication = (WebApplication) source;
+            // if they are hosted on the same compute --> we can set the compute private address to
+            Compute computeMysqlDatabase = getCompute(mysqlDatabase);
+            Compute computeWebApplication = getCompute(webApplication);
+            if (computeMysqlDatabase.equals(computeWebApplication)) {
```
hnicke: What happens if that is not the case?

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
```diff
@@ -45,85 +54,118 @@
...
-            CFNInit init = new CFNInit(CONFIG_SETS);
-            cfnModule.putCFNInit(nodeName, init);
-            //takes default 8GB storage but volume
-            cfnModule.resource(Instance.class, nodeName)
-                .keyName(cfnModule.getKeyNameVar())
-                .securityGroupIds(webServerSecurityGroup)
-                .imageId(imageId)
-                .instanceType(instanceType);
+            logger.info("Visit Compute node '{}'.", node.getEntityName());
+            if (cfnModule.checkComputeToEc2(node)) {
```
mmunozba: Maybe I'm misunderstanding this, but doesn't this mean that a Compute node only gets transformed if an Apache is hosted on it? Since `addComputeToEc2` only gets called in the `visit()` method of Apache in `PrepareModelNodeVisitor`.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelRelationshipVisitor.java
```diff
@@ -0,0 +1,60 @@
...
+        this.logger = context.getLogger(getClass());
+        this.topology = context.getModel().getTopology();
+    }
+
+    @Override
+    public void visit(ConnectsTo relation) {
+        logger.info("Prepare ConnectsTo relation '{}'.", relation.getEntityName());
+        RootNode source = topology.getEdgeSource(relation);
+        RootNode target = topology.getEdgeTarget(relation);
+        if (source instanceof WebApplication && target instanceof MysqlDatabase) {
```
hnicke: What happens if that is not the case? Maybe log that you drop the relationship.

server/src/test/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationLifecycleTest.java
```diff
@@ -29,12 +40,26 @@
 
     @Before
     public void setUp() throws Exception {
-        PluginFileAccess access = new PluginFileAccess(new File(""), tmpdir, mock(Log.class));
-        EffectiveModel effectiveModel = new EffectiveModelFactory().create(TestCsars.VALID_MINIMAL_DOCKER_TEMPLATE, logMock());
-
+        PluginFileAccess accessL = new PluginFileAccess(
+            new File(tmpdir, "sourceDir"),
+            new File(tmpdir, "targetDir"),
+            mock(Log.class));
```
hnicke: you could simply use `logMock()` instead of `mock(Log.class)`. The calls to getLogger() are already stubbed.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
```diff
@@ -296,4 +338,18 @@ private String getFileURL(String bucketName, String objectKey) {
     public CapabilityMapper createCapabilityMapper() {
         return new CapabilityMapper(cfnModule.getAWSRegion(), cfnModule.getAwsCredentials(), logger);
     }
+
+    private Set<Compute> getHostsOfConnectedTo(RootNode node) {
```
mmunozba: Maybe extract this method? See above.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelNodeVisitor.java
```diff
@@ -0,0 +1,109 @@
...
+        logger.info("Prepare MysqlDatabase node '{}'.", node.getEntityName());
+        //if certain values aren't given, fill them
+        if (node.getPassword().isPresent()) {
+            //password needs to be at least 8 characters long
+            String password = node.getPassword().get();
+            if (password.length() < minPWLength) {
+                logger.warn("Database password to short, creating new random password");
```
mmunozba: to -> too

server/src/test/resources/csars/yaml/valid/lamp-noinput/my_app/myphpapp.php
```diff
@@ -10,6 +10,19 @@
 }
 
 
+/**
```
hnicke: Please also rezip the corresponding csar (run csar-make)

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationModule.java
```diff
@@ -75,6 +83,7 @@ public CloudFormationModule(PluginFileAccess fileAccess, String awsRegion, AWSCr
             (KEYNAME_CONSTRAINT_DESCRIPTION);
         keyNameVar = template.ref(KEYNAME);
         cfnInitMap = new HashMap<>();
+        computeToEc2 = new HashSet<>();
```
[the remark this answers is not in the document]
mklopp: We don't want to transform every compute node to an ec2 instance. For example if an database is the only node hosted on an compute we dont want to transform it.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
mklopp: Yeah this is wrong, I changed it

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CloudFormationNodeVisitor.java
mklopp: I extracted these into a parent class

PR #436, opened 2018-01-24
Add Deployment phase to TransformationLifecycle

### LINKED ISSUE #438: Add Deployment phase to Transformation Lifecycle
We want to have a seperate deployment phase that can be executed if the plugin supports this feature and the user has selected it!

c-mueller:
This PR implements a Lifecycle phase in the TransformationLifecycle. This is intended to allow the automatic Deployment from within the TOSCAna Instance.

Currently the deployment step has to be selected before a Transformation is started. You cannot deploy on the instance if it was not selected in the first place. 

Code Changes:
- The Classes and Interfaces `TransformationPlugin` and `AbstractPlugin` have been removed. Their Functionality is now in the `TOSCAnaPlugin` class. This Class is formerly known as `LifecycleAwarePlugin`.
    - This Change does not affect any plugin because the original idea behind this (completely ignoring the Lifecycle to write a Plugin) is no longer of interest.
- The Lifecycle now has a `deploy` method representing the Deployment operation. 
    - The Common implementation of this will Follow in another PR. Please Dont Implement this method yet.
- The flag defining that in-app deployment is supported gets stored in the `Platform` class. For this a new Constructor `Platform(String, String, boolean, Set<SimpleProperty>)` has been added. The old Constructors still exist but the deployment flag is always set to `false`
- If a plugin supports Deployment the Platform class will automatically add a Property with the key `deploy_after_transformation`. This property is a Boolean property and the default value is `false`. 
- The `TransformationContext` has a method `shouldDeploy()` that returns true if the property above is set. If this method returns `true` the lifecycle will execute the deploy operation.
- Property Validators now return `false` if the input is null
    - This might result in some unintended sideeffects. Reverting this should not be a issue.

Resolves #438


server/src/main/java/org/opentosca/toscana/core/plugin/TOSCAnaPlugin.java
```diff
@@ -1,36 +1,45 @@
...
      executed by the transform method.
      <p>
      The tasks get stored in a list (environment tasks)
      */
-    public LifecycleAwarePlugin(Platform platform) {
-        super(platform);
+    public TOSCAnaPlugin(Platform platform) {
+        this.platform = platform;
+        if (platform == null) {
+            throw new IllegalStateException("Platform must not be null");
```
hnicke: Could be simplified like this:
this.platform = Objects.requireNonNull(platform);

server/src/main/java/org/opentosca/toscana/core/transformation/TransformationContext.java
```diff
@@ -43,4 +44,13 @@ public Logger getLogger(Class clazz) {
     public PluginFileAccess getPluginFileAccess() {
         return access;
     }
+
+    public boolean performDeployment() {
```
hnicke: maybe hide this method ; plugin implementions should not see this method.

Issue #413, opened 2018-01-14
Include parameters in the deployment for cloudformation templates


mmunozba:
Currently, the CloudFormation Plugin only adds a command to deploy the template without setting values for the needed additional parameters in the template. This could be done by adding the --parameter-overrides followed by the parameter key value pairs. (See the [AWS CLI Reference](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/deploy/index.html) for more information)

Without these parameters set, the template cannot be deployed, rendering the current deploy command useless.
We must build a way to get these parameters from the user both when the user deploys the template by himself as well as when he instructs TOSCAna to do so.

PR #442, opened 2018-01-26
Include parameters in the deployment for cloudformation templates

mmunozba:
From #413.


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -82,10 +84,25 @@ private void createDeployScript() throws IOException {
         logger.debug("Creating deploy script.");
         BashScript deployScript = new BashScript(cfnModule.getFileAccess(), FILENAME_DEPLOY);
         deployScript.append(EnvironmentCheck.checkEnvironment("aws"));
+        deployScript.append("source " + FILENAME_UPLOAD + ".sh");
```
mklopp: Doesn't this need a check whether this file is/will be created? Because if there are no files to be uploaded the `file-upload.sh` won't be created and this should throw an error.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -82,10 +84,25 @@ private void createDeployScript() throws IOException {
...
-            + CLI_PARAM_STACKNAME + cfnModule.getStackName() + " "
-            + CLI_PARAM_TEMPLATEFILE + TEMPLATE_YAML);
+        StringBuilder deployCommand = new StringBuilder("");
+        deployCommand.append(CLI_COMMAND_CREATESTACK + CLI_PARAM_STACKNAME)
+            .append(cfnModule.getStackName()).append(" ")
+            .append(CLI_PARAM_TEMPLATEFILE).append(TEMPLATE_YAML);
+
+        // Add parameters if needed
+        Map<String, Parameter> parameters = cfnModule.getParameters();
+        if (!CollectionUtils.isEmpty(parameters)) {
```
mklopp: Whats the difference to `!parameters.isEmpty()` ?

mmunozba: This also provides safety for null checks, so if the collection is null, there is no NullPointerException. That cannot happen here (I think) so `!parameters.isEmpty()` would, as you suggested, suffice.

I just personally like using CollectionUtils.isEmpty() in general, so I don't run into unexpected NullPointerExceptions.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
mmunozba: You're right, I didn't think of this case. I'm going to have to check how ``source file-upload.sh` acts when the file does not exist and probably add a check before that line.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
hnicke: Often though, crashing early on encountering a nullpointer is better than propagating the error even further...

PR #433, opened 2018-01-22
Rework Transformation Flow for Cloud Foundry

orferch:
See #427

PR #448, opened 2018-01-29
[WIP] Adapt NodeVisitor to recognize services

JensMueller2709:
In the transformation rework #433 we create own stack for each top node (e.g.WebApplication). Each stack is a cloud foundry application with the exception of a service node e.g. a mysql-database. 
The service belongs to an application. Therefore we have to find the application which the service belongs to.


JensMueller2709:
merged into PR #433

PR #459, opened 2018-01-30
Fix Logger File Overflow

c-mueller:


server/src/integration/java/org/opentosca/toscana/core/LogOverflowIT.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.File;
+
+import org.opentosca.toscana.core.transformation.logging.Log;
+import org.opentosca.toscana.core.transformation.logging.LogImpl;
+
+import org.junit.Test;
+import org.slf4j.Logger;
+
+@SuppressWarnings("Duplicates")
+public class LogOverflowIT extends BaseIntegrationTest {
```
hnicke: Why is this not a unit test?

c-mueller: Potential timeout issues. Took about 35 Seconds when running the first time.

PR #463, opened 2018-01-31
[WIP] Add Samples Repository as Submodule

c-mueller:
This will add the sample repository as a git submodule. 

I will try to copy all versions of the LAMP CSAR over in the other repo and test it afterwards, thats also why this is still WIP


c-mueller:
The content of the repo will be copied over.

PR #467, opened 2018-01-31
Update Property Setting Response

c-mueller:
Changes the Property response when setting a property from:
```json
{
    "valid_inputs": {
         "unsigned_integer": false
     }
}
```

to

```json
{
    "_links": {
        "self": {
            "href": "http://localhost:8080/csars/mongo-db/transformations/p-a/properties"
        }
    },
    "properties": [
        {
            "valid": false,
            "key": "unsigned_integer",
            "type": "unsigned_integer",
            "description": "",
            "required": true,
            "value": "-111"
        }
    ]
}
```


server/src/main/java/org/opentosca/toscana/api/TransformationController.java
```diff
@@ -778,18 +778,16 @@ This Mapping is used to post the inputs (properties) to the server. this will se
...
-//            transformation.setState(TransformationState.READY);
-//        }
         if (!somethingFailed) {
             return ResponseEntity.ok().build();
         } else {
-            return ResponseEntity.status(HttpStatus.NOT_ACCEPTABLE).body(new SetPropertiesResponse(successes));
+            SetPropertiesResponse response = new SetPropertiesResponse(
+                csarId,
+                platformId,
+                setPropertiesRequest.getProperties(),
```
hnicke: I think this will lead to faulty behaviour in combination with default values.

Imagine a SET properties request containing this property:
```yaml
property:
  value: null
  required: true
  default: "my-default"
  type: text
```

What currently happens: You send the same property back (including success indicator)

But given the fact that the property is required and has a default value, following behaviour will be observable:
```java
property.setValue(null);
assertEquals(property.getDefault(), property.getValue());
assertNull(property.getValueWithoutDefault);
```

The api response should also reflect this evaluation.
Not sure whether it's still appropriate to not have default values in the `PropertyWrap` class.

server/src/main/java/org/opentosca/toscana/api/model/PropertyWrap.java
```diff
@@ -42,7 +42,8 @@ public String getKey() {
     @ApiModelProperty(
         required = true,
         notes = "The \"Datatype\" of a property. Supported Types are \"name\", \"text\", \"boolean\", " +
-            "\"integer\", \"unsigned_integer\", \"float\" and \"secret\"",
+            "\"integer\", \"unsigned_integer\", \"float\" and \"secret\". If you try to set a value of a non existant key." +
```
hnicke: You might use single quote characters for quoting inside java string, makes these strings much more readable.

PR #469, opened 2018-02-01
Allow EC2 instances to access S3Buckets

### LINKED ISSUE #444: CloudFormation can't access bucket files
Currently, when CloudFormation attempts to deploy the template created by our Plugin, CFN-init cannot access the files in the S3 bucket which are specified in our template.
This needs to be fixed in order to allow successful deployment.

Links that could help with this:
- https://docs.aws.amazon.com/de_de/AWSCloudFormation/latest/UserGuide/aws-resource-authentication.html
- https://aws.amazon.com/de/blogs/devops/authenticated-file-downloads-with-cloudformation/
- https://docs.aws.amazon.com/AmazonS3/latest/dev/PresignedUrlUploadObjectJavaSDK.html
- https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html

mmunozba:
Fixes #444.


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationModule.java
```diff
@@ -110,7 +131,15 @@ public CFNInit getCFNInit(String resource) {
     public void putFileToBeUploaded(String filePath) {
         this.filesToBeUploaded.add(filePath);
     }
-
+    
+    public List<String> getAuthenticationList() {
+        return authenticationList;
+    }
+    
+    public void putAuthentication(String instanceName) {
```
mklopp: should be `addAuthentication`. `put` is for maps and sets

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/TransformModelNodeVisitor.java
```diff
@@ -213,9 +213,11 @@ private void handleOperation(Operation operation, String serverName, String conf
         for (String dependency : operation.getDependencies()) {
             String cfnSource = getFileURL(cfnModule.getBucketName(), dependency);
 
-            logger.debug("Marking " + dependency + " as file to be uploaded.");
+            logger.debug("Marking '{}' as file to be uploaded.", dependency);
             cfnModule.putFileToBeUploaded(dependency);
-
+            logger.debug("Marking '{}' as instance in need of authentication.", serverName);
+            cfnModule.putAuthentication(serverName);
```
mklopp: Maybe check first if the `serverName` is already maked for authentication. But I guess it is a set so it doesn't matter if you add it multiple times

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/TransformModelNodeVisitor.java
```diff
@@ -233,8 +235,10 @@ private void handleOperation(Operation operation, String serverName, String conf
             String artifact = operation.getArtifact().get().getFilePath();
             String cfnSource = getFileURL(cfnModule.getBucketName(), artifact);
 
-            logger.debug("Marking " + artifact + " as file to be uploaded.");
+            logger.debug("Marking '{}' as file to be uploaded.", artifact);
             cfnModule.putFileToBeUploaded(artifact);
+            logger.debug("Marking '{}' as instance in need of authentication.", serverName);
+            cfnModule.putAuthentication(serverName);
```
mklopp: Same as above

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/TransformModelNodeVisitor.java
mmunozba: You're right, even if it doesn't matter, it might be good to know for debugging reasons.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationModule.java
```diff
@@ -264,9 +285,65 @@ private String getRandomStackName() {
         return "toscana-stack-" + UUID.randomUUID();
     }
 
+    /**
+     Returns the <tt>Authentication<tt> to access S3 for this module.
```
hnicke: This class is too big. Please extract some aspects into their own class. E.g., this whole authentication aspect can get extracted.

mmunozba: I extracted some methods that are responsible for similar things into their own utility classes.

PR #480, opened 2018-02-02
Add Express Application to Template

c-mueller:


server/src/test/resources/csars/yaml/valid/express/template.yml
```diff
@@ -31,6 +31,10 @@ topology_template:
           start:
             implementation:
               primary: simpletaskapp_js/start_stajs.sh
+            inputs:
+                REDIS_HOST: { get_attribute: [ redis_compute, private_address ] }
```
hnicke: Not sure if snakeyaml would complain, but these entries have wrong indentation

c-mueller: Intellij was a bit stupid there

hnicke:
With the latest commit, the winery validation passes.

PR #484, opened 2018-02-03
Disable autodeletion of CSAR after upload when validation fails

### LINKED ISSUE #483: Disable automatic csar deletion after failed validation
Currently, when uploading a CSAR and validation fails, the CSAR is automatically deleted.
This behaviour shall be disabled.

hnicke:
Resolves #483


hnicke:
Travis failed because the Base Image Mapper experienced a SocketTimeOut exception while trying to communicate with docker hub..

server/src/main/java/org/opentosca/toscana/api/CsarController.java
```diff
@@ -160,14 +153,9 @@ This Request (Supporting Post and Put methods) uploads a csar to the transformer
     @ApiResponses({
         @ApiResponse(
             code = 201,
-            message = "The upload and parsing of the csar was sucessful",
+            message = "The upload of the csar was successful",
             response = Void.class
         ),
-        @ApiResponse(
-            code = 400,
```
c-mueller: is this Method only lauching a upload job now? If not: what is returned if the upload fails?

hnicke: Csar upload now will return 201 even if csar validation fails. In the old implementation, code 400 would only get returned if the csar validation failed. I eliminated this possibility, so I also removed return code 400.

c-mueller: Okay.

PR #497, opened 2018-02-04
Enable general logging to file

hnicke:
Enables logging to file (/var/log/toscana.log).
Make sure you can write to /var/log.

### DIFF
--- a/docker/alpine-dind/toscana-dind-entrypoint.sh
+++ b/docker/alpine-dind/toscana-dind-entrypoint.sh
@@ -1,3 +1,6 @@
 #!/bin/bash
 dockerd &
-java -jar server.jar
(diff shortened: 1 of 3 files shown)


docker/alpine-dind/toscana-dind-entrypoint.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
 dockerd &
-java -jar server.jar
+logfile=/var/log/toscana.log
```
c-mueller: I would consider this a Environment Variable (in the dockerfile) `ENV LOGFILE=/var/log/toscana.log` To allow external customisation. 

This would have to be added in the dockerfile for the jenkins deployment and the one used by dockerhub

PR #498, opened 2018-02-04
Fix Connections within the same Pod

c-mueller:
Connections of containers in the same pod now work properly again.

### DIFF
--- a/server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/MapperConstants.java
+++ b/server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/MapperConstants.java
@@ -12,7 +12,7 @@ import static org.opentosca.toscana.model.capability.OsCapability.Architecture.x
 public class MapperConstants {
     public static final String DEFAULT_IMAGE_PATH = "library/ubuntu:latest";
     public static final String DEFAULT_IMAGE_DISTRO = "ubuntu";
-    public static final String DOCKER_HUB_URL = "https://registry.hub.docker.com";
+    public static final String DOCKER_HUB_URL = "";
     public static final Map<Architecture, String> ARCHITECTURE_MAP;
 
     static {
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/opentosca/toscana/plugins/kubernetes/docker/mapper/MapperConstants.java
```diff
@@ -12,7 +12,7 @@
 public class MapperConstants {
     public static final String DEFAULT_IMAGE_PATH = "library/ubuntu:latest";
     public static final String DEFAULT_IMAGE_DISTRO = "ubuntu";
-    public static final String DOCKER_HUB_URL = "https://registry.hub.docker.com";
+    public static final String DOCKER_HUB_URL = "";
```
hnicke: You intend to change the default value of the platform property to an empty string, isn't it? This feels hacky. You might rather assign an empty String directly as the platform property default value.

server/src/main/java/org/opentosca/toscana/plugins/kubernetes/visitor/imgtransform/DockerfileBuildingVisitor.java
```diff
@@ -170,7 +174,31 @@ private void handleDefault(RootNode node) {
...
+            //The same compute node
+            node.getRequirements().forEach(e -> {
+                if (e.getRelationship().isPresent() && e.getRelationship().get() instanceof ConnectsTo) {
+                    for (Object o : e.getFulfillers()) {
+                        if (o instanceof RootNode) {
+                            NodeStack targetStack = this.connectionGraph.vertexSet()
+                                .stream().filter(ek -> ek.hasNode((RootNode) o)).findFirst().orElse(null);
+                            if (targetStack != null &&
+                                targetStack.getComputeNode() == this.stack.getComputeNode()) {
+                                adresses.put(this.stack, this.stack.getComputeNode().getPrivateAddress().get());
```
hnicke: How do you know that the Optional is present? You might want to replace the `get()` with `getOrElse(null)`.

c-mueller: I Set The Private address of every compute node beforehand. (After the initial grouping in pods)

PR #505, opened 2018-02-05
Dedicated phase instances for each transformation

hnicke:
Currently, lifecycle phases belong to the plugins. The problem is that multiple transformations (e.g., all transformations of plugin 'kubernetes') share the same phase instances. This results in wrong phase states.

This patch moves the lifecycle phases from the plugin instances to the Lifecycle classes (one lifecycle instance is created per transformation). Now every transformation has its own set of phases.


server/src/main/java/org/opentosca/toscana/core/plugin/TOSCAnaPlugin.java
```diff
@@ -1,155 +0,0 @@
...
-import org.opentosca.toscana.core.util.LifecyclePhase;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-/**
- This class represents the base class for every Plugin that wants to implement the Transformation lifecycle
-
- @param <LifecycleT> The class type of the plugin specific implementation of the Lifecycle interface */
-public abstract class TOSCAnaPlugin<LifecycleT extends TransformationLifecycle> implements Lifecycle {
```
c-mueller: I honestly dont like this whole renaming thing. The original file was written by me. even though you changed it many lines would still be mine. After that ALL lines belong to you and i honestly dont like that. Als why should we rename this? Just because of some very "important" naming convention?
image

hnicke: I did not intend to 'steal' lines. I simply keep refactoring code all the time.
I'm trying to roll back the changes..

hnicke: In addition to renaming, the file got moved in order to make use of package visibility. This caused the change of the line ownership. git-blame's `-C` flag solves your issue. Is that ok? We have to keep on refactoring the codebase...

PR #516, opened 2018-02-09
Implement support for JavaRuntime and JavaApplication

### LINKED ISSUE #515: Implement support for JavaApplication and JavaRuntime type

c-mueller:
Resolves #515


server/src/main/java/org/opentosca/toscana/plugins/kubernetes/visitor/imgtransform/DockerfileBuildingVisitor.java
```diff
@@ -76,6 +78,33 @@ public void visit(Nodejs node) {
...
+        handleDefault(node);
+    }
+
+    @Override
+    public void visit(JavaApplication node) {
+        handleDefault(node);
+        //Copy over jar and add command to entrypoint
+        String jarPath = node.getJar().getFilePath();
+        try {
+            builder.copyFromCsar(jarPath, "", node.getEntityName() + ".jar");
```
hnicke: In my opinion, not using the original jar name, but `node.getEntityName() + ".jar"`, feels unintuitive. Why should we rename artifacts /scripts? This is quite confusing for the user.

c-mueller: Had issues with that originaly. because the files get copied in the root directory `/toscana-root` and im not creating any subfolders there.

I might consider changing this behaviour in another issue (is probably too big for a quick fix)

PR #518, opened 2018-02-09
Database environment variables recognition

JensMueller2709:


server/src/integration/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLampIT.java
```diff
@@ -48,8 +48,8 @@ protected EffectiveModel getModel() throws Exception {
 
     @Override
     protected void onSuccess(File outputDir) throws Exception {
-        //System.out.println("You can stop me now");
-        //Thread.sleep(5000);
+        System.out.println("You can stop me now");
```
hnicke: Looks like this should get removed before merging.

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/CloudFoundryLifecycle.java
```diff
@@ -44,6 +45,13 @@ public boolean checkModel() {
 
     @Override
     public void prepare() throws NoSuchPropertyException {
+
+        //just for testing in this branch
```
hnicke: Still 'just for testing'?

server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/visitors/PrepareVisitor.java
```diff
@@ -0,0 +1,33 @@
...
+    private final Logger logger;
+
+    public PrepareVisitor(Logger logger) {
+        this.logger = logger;
+    }
+
+    @Override
+    public void visit(MysqlDatabase node) {
+        logger.debug("Prepare Model. Set placeholders for mysql database node");
+        node.setUser(CF_PREPARE_DB_PLACEHOLDER_USER);
```
hnicke: What happens in case of 2 MysqlDatabase node? I thought the placeholders need to be unique for the algorithm to work.

Issue #406, opened 2018-01-12
Check which Mode, Owner Group are needed for Cloudformation Files


mmunozba:
Currently the handleOperation() function always sets the cfnFileMode, cfnFileOwner and cfnFileGroup to "000644", "root" and "root" respectively. This might need/should be set dynamically in the future.

PR #519, opened 2018-02-09
Support Express Model

mmunozba:
- [x] Support Nodejs node type
- [x] Support redis database


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CheckModelRelationshipVisitor.java
```diff
@@ -28,7 +29,8 @@ public CheckModelRelationshipVisitor(TransformationContext context) {
     public void visit(ConnectsTo relation) {
         RootNode source = topology.getEdgeSource(relation);
         RootNode target = topology.getEdgeTarget(relation);
-        if (!(source instanceof WebApplication && target instanceof MysqlDatabase)) {
+        if (!(source instanceof WebApplication && target instanceof MysqlDatabase) &&
```
hnicke: If the second condition is true, the first condition is also true. You might want to omit the first statement.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -197,7 +193,6 @@ private String uploadFile(String objectKey, String filePath) {
     public void copyUtilScripts() throws IOException {
         String resourcePath = "/cloudformation/scripts/util/";
         String outputPath = "output/scripts/util/";
```
JensMueller2709: You could use the static string from the AbstractLifecycle `public static final String UTIL_DIR_PATH`

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -206,14 +201,49 @@ public void copyUtilScripts() throws IOException {
...
+        copyUtilFile(utilScripts, resourcePath, outputPath);
+    }
+
+    /**
+     Copies all needed cloudformation utility files to the target artifact.
+     Note: Theses are the files that actually need to be uploaded and accessed by EC2 instances unlike the util scripts.
+     */
+    public void copyUtilDependencies() throws IOException {
+        String resourcePath = "/cloudformation/files/util/";
+        String outputPath = "output/files/";
```
JensMueller2709: see above with `public static final String OUTPUT_DIR`

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/TransformModelNodeVisitor.java
```diff
@@ -143,8 +168,30 @@ public void visit(MysqlDatabase node) {
...
+            String computeHostName = toAlphanumerical(computeHost.getEntityName());
+            operationHandler.handleGenericHostedNode(node, computeHost);
+
+            //Open Dbms port
+            String SecurityGroupName = computeHostName + SECURITY_GROUP;
+            SecurityGroup securityGroup = (SecurityGroup) cfnModule.getResource(SecurityGroupName);
+            if (node.getPort().isPresent()) {
+                Integer dbmsPort = node.getPort().orElseThrow(() -> new IllegalArgumentException("Database " +
+                    "port not set"));
+                securityGroup.ingress(ingress -> ingress.cidrIp("0.0.0.0/0"), "tcp", dbmsPort);
```
JensMueller2709: Consider making static strings for `"0.0.0.0/0"` and maybe `tcp`

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/TransformModelNodeVisitor.java
```diff
@@ -193,108 +236,41 @@ public void visit(WebApplication node) {
...
+            operationHandler.handleConfigure(node, computeHostName);
+            //handle start
+            operationHandler.handleStart(node, computeHostName);
+            //add NodeJs create script
+            operationHandler.addCreate(FILEPATH_NODEJS_CREATE, computeHostName);            
+            
+            //Open port 3000
+            String SecurityGroupName = computeHostName + SECURITY_GROUP;
+            SecurityGroup securityGroup = (SecurityGroup) cfnModule.getResource(SecurityGroupName);
+            securityGroup.ingress(ingress -> ingress.cidrIp("0.0.0.0/0"), "tcp", 3000);
```
JensMueller2709: see above

server/src/main/resources/cloudformation/files/util/create-nodejs.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
+sudo apt-get update
```
JensMueller2709: just a question: do you only want to update or maybe upgrade as well?
if not forget the question

c-mueller: The reason why you run update here is to Download the "Package Database" for node js that has been added using the command above the update command.

There is no intention to run a full system upgrade (At least there was none when i wrote this script for the express model) We just want to fetch the latest version of the database to install the latest version of nodejs

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -206,14 +201,49 @@ public void copyUtilScripts() throws IOException {
...
-            if (!utilScript.isEmpty()) {
+        copyUtilFile(utilScripts, resourcePath, outputPath);
+    }
+
+    /**
+     Copies all needed cloudformation utility files to the target artifact.
+     Note: Theses are the files that actually need to be uploaded and accessed by EC2 instances unlike the util scripts.
+     */
+    public void copyUtilDependencies() throws IOException {
+        String resourcePath = "/cloudformation/files/util/";
```
c-mueller: Also consider making this constant

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -206,14 +201,49 @@ public void copyUtilScripts() throws IOException {
...
+        });
+    }
+    
+    public void copyUtilFile(List<String> files, String resourcePath, String outputPath) throws IOException {
+        PluginFileAccess fileAccess = cfnModule.getFileAccess();
+        for (String file : files) {
+            if (!file.isEmpty()) {
                 //Copy the file into the desired directory
-                logger.debug("Adding " + utilScript + " to the target artifact.");
+                logger.debug("Adding " + file + " to the target artifact.");
```
c-mueller: Please consider using `logger.debug("Adding {} to the target artifact.", file);`

This will only concatenate if the loglevel requires the rendering of the message (i.e. level is set to debug or trace)

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/util/OperationHandler.java
```diff
@@ -0,0 +1,242 @@
...
+        } else {
+            logger.debug("'{}' already marked as instance in need of authentication. " +
+                "Skipping authentication marking.", serverName);
+        }
+        cfnModule.putAuthentication(serverName);
+
+        return new CFNFile(ABSOLUTE_FILE_PATH + filePath)
+            .setSource(cfnSource)
+            .setMode(mode) //TODO Check what mode is needed (read? + execute?)
+            .setOwner(OWNER_GROUP_ROOT) //TODO Check what Owner is needed
```
c-mueller: Are these todos still a thing? please consider opening issues for them if they wont be resolved in this PR.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/util/StackUtils.java
```diff
@@ -85,4 +85,16 @@ public static Fn getUserDataFn(String resource, String configsets, CloudFormatio
...
+    /**
+     Returns the URL to the file in the given S3Bucket.
+     e.g. http://bucketName.s3.amazonaws.com/objectKey
+
+     @param bucketName name of the bucket containing the file
+     @param objectKey  key belonging to the file in the bucket
+     @return URL for the file
+     */
+    public static String getFileURL(String bucketName, String objectKey) {
+        return CloudFormationModule.URL_HTTP + bucketName + CloudFormationModule.URL_S3_AMAZONAWS + "/" + objectKey;
```
c-mueller: You could statically import the CloudFormationModule class. to reduce line length,

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/CheckModelNodeVisitor.java
```diff
@@ -55,19 +58,34 @@ public void visit(Compute node) {
...
+    public void visit(Dbms dbms) {
+        // noop
+    }
+    
     @Override
     public void visit(MysqlDbms node) {
         // noop
     }
-
+
```
c-mueller: Please run a reformat with the proper settings to resolve such issues. (empty lines should not contain spaces)

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/TransformModelNodeVisitor.java
```diff
@@ -193,108 +236,41 @@ public void visit(WebApplication node) {
...
+            String computeHostName = toAlphanumerical(computeHost.getEntityName());
+            
+            //handle configure
+            operationHandler.handleConfigure(node, computeHostName);
+            //handle start
+            operationHandler.handleStart(node, computeHostName);
+            //add NodeJs create script
+            operationHandler.addCreate(FILEPATH_NODEJS_CREATE, computeHostName);            
+            
+            //Open port 3000
```
c-mueller: Please consider reading the port from the node, instead of hardcoding port 3000. 
This could be achieved like this (Taken from the DockerfileBuildingVisitor):
```java
node.getCapabilities().forEach(e -> {
 try {
   if (e instanceof EndpointCapability) {
     if (((EndpointCapability) e).getPort().isPresent()) {
       ports.add(((EndpointCapability) e).getPort().get().port);
     }
   }
 } catch (Exception ex) {
   logger.warn("Failed reading Port from node {}", node.getEntityName(), ex);
 }
});
```

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/util/OperationHandler.java
mmunozba: See #406.

mmunozba:
Unfortunately, the start script still doesn't work properly. I've exhausted a bunch of different solutions, but I didn't manage to find the cause. Since I've already spent a lot of time on trying to find the cause without coming up with results, I've decided to postpone it to a later day (I'll create an issue soon).
The start script can be run manually (although for some reason that also doesn't work 100% of the time) but the automatic deployment fails every time.

Since I'm currently still working alone on the CloudFormation plugin, I want to shift my attention to other things for now (environment variables, java model) and get those working first, since they will probably be more manageable. 
But since I need some of the changes of this PR, I'd like to merge this right now and patch the start script problems at a later date.

PR #531, opened 2018-02-10
Core implementation of the WebApp

nfode:
The WebApp might not be finished, it has some "rough edges" and some bugs but the core functionality is there. At the moment it is hard to manage open issues and bugs becaues this PR is really huge so I propose to merge this. If someone with basic angular understanding can review this today it would be nice.


nfode:
@hnick the 300+ codacy issues came from the single quotemark guideline i use, it wantet double quotemarks.

PR #532, opened 2018-02-11
Improve script creation

mmunozba:


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -141,6 +126,54 @@ private void createFileUploadScript() throws IOException {
         }
     }
 
+    /**
+     Creates the script for creating the CloudFormation stack from the template.
+     */
+    private void createCreateStackScript() throws IOException {
```
c-mueller: Maybe consider renaming this to `writeStackCreationScript`

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -141,6 +126,54 @@ private void createFileUploadScript() throws IOException {
...
+
+        deployCommand.append(" &");
+        createStackScript.append(deployCommand.toString());
+    }
+
+    /**
+     Creates a script to delete the S3Bucket and the deployed Stack.
+     Note: May or may not be included in the final version. Currently used for quicker manual debugging.
+     */
+    private void createCleanUpScript() throws IOException {
```
c-mueller: Simmilar to above. `create` sounds a bit wrong, because you actually write/copy these files in the working directory

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -141,6 +126,54 @@ private void createFileUploadScript() throws IOException {
...
+        if (!parameters.isEmpty()) {
+            logger.debug("Adding parameters to create stack command.");
+            deployCommand.append(" " + CLI_PARAM_PARAMOVERRIDES);
+            for (Map.Entry<String, Parameter> entry : parameters.entrySet()) {
+                String id = entry.getKey();
+                deployCommand.append(" ").append(id).append("=$").append(id).append("Var");
+            }
+        }
+
+        deployCommand.append(" &");
```
c-mueller: Assuming this is part of a Bash/Shell script. Why do you want background ("parallel") execution of this command?

mmunozba: This was done because the `aws deploy` command creates a stack that usually takes about 15 min to complete. During this period, nothing happens besides the message `Creating changeset...` and a single confirmation of the successful creation of the stack. Both these messages and any additional messages (e.g. errors, warnings, etc.) still get shown on the console even if the command runs in the background.

So, instead of having the console blocked for 15 min, @mklopp suggested running the command in the background instead. I'm not sure if this is something we should change for the final version but if there are reasons to not do this, then I'll be happy to consider them.

nfode: I suggest removing the `&`, it is highly anoying using a shell that randomly outputs logs. If I want to run the script in the background I can add it  to my higher level command: `./deploy.sh &`.

nfode: I suggest removing the `&`, it is highly anoying using a shell that randomly outputs logs. If I want to run the script in the background I can add it  to my higher level command: `./deploy.sh &`.

c-mueller: Thats not the only problem with this. I absolutely dont like the idea of a background operation that performs the deployment. 

The problem with the blocked console is also easy to solve and should not be our problem. If the end user is incapable of creating a new Tab in the terminal or using tools like `screen` or `tmux` we cannot help him anyways...

mmunozba:
Gonna go ahead and merge this since I've addressed all requested changes.

PR #551, opened 2018-02-14
Support Java application

JensMueller2709:


server/src/main/java/org/opentosca/toscana/plugins/cloudfoundry/filecreator/ApplicationHandler.java
```diff
@@ -81,7 +81,7 @@ private void copyData(Application unrealApplication, Application parentApplicati
         unrealApplication.getExecuteCommands().forEach((k, v) -> parentApplication.addExecuteFile(k, v));
 
         //copy mysql configure file
-        unrealApplication.getConfigMysql().forEach(sqlFile -> parentApplication.addConfigMysql(sqlFile));
+        unrealApplication.getConfigMysql().forEach((service, sqlFile) -> parentApplication.addConfigMysql(service, sqlFile));
```
mmunozba: According to Intellij:
This line:
`unrealApplication.getConfigMysql().forEach((service, sqlFile) -> parentApplication.addConfigMysql(service, sqlFile));`
Can be replaced with:
`unrealApplication.getConfigMysql().forEach(parentApplication::addConfigMysql);`

Same for all the other lambda expressions in this class.

PR #556, opened 2018-02-15
Enhance input form with description and checkbox

nfode:


app/src/app/components/input/input.component.html
```diff
@@ -16,10 +16,18 @@
         <tr *ngFor="let item of inputs; let n=index">
             <td>{{item.key}}</td>
             <td>{{item.type}}</td>
-            <td><input [value]="item.value"  (input)="change(item, $event.target.value)" name="item.key"
-                       [ngClass]="{'invalid': !item.valid, 'valid':item.valid}"/>
+            <td *ngIf="!hasTypeBoolean(item)"><input [value]="item.value"
```
c-mueller: Please consider Reformatting this.

app/src/app/components/input/input.component.scss
```diff
@@ -42,4 +42,14 @@ form {
             margin: 2px;
         }
     }
+    input[type=checkbox] {
```
c-mueller: Why does github mark this red?

Are you sure this is syntactically correct?

nfode: It is correct. this is a so called attribute selector. [Check this link.](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors)

c-mueller: Okay. Was just wondering why this is highlighted red

PR #568, opened 2018-02-18
Refactor KubernetesLifecycle

### LINKED ISSUE #566: Refactor KubernetesLifecycle

c-mueller:
This PR moves most of the Transformation logic implemented in the KubernetesLifecycle into seperate Classes. 

Resolves #566


c-mueller:
@JensMueller2709 I think the Generalisation will be part of another PR we also have to consider how relevant the Generalisation is for you. If it doesn't affect you iwould not consider it "high priority"

PR #569, opened 2018-02-18
Fix Dockerfile not exposing Ports and build E2E Integration Test for registry Pushing

### LINKED ISSUE #596: Ports don't get exposed in Dockerfile

c-mueller:
To Quote my commit message: 
```
The Kubernetes Plugin uses this CSAR for testing the Pushing capability
of the Kubernetes lifecycle.
The CSAR represents a Staticly linked linux binary that runs a HTTP
server on Port 8080 showing some system Information (Similar to the
        Docker Image 'tutum/hello-world')
The Source Code of the 'App' can be Found in the samples Repository
```

This test tries to transform the image, push it to a local registry, pull it from there (after removing the image localy) and running the image within docker.

This PR also resolves #596

### DIFF
--- a/server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/model/KubernetesPushingGopherIT.java
+++ b/server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/model/KubernetesPushingGopherIT.java
@@ -1,0 +1,155 @@
+package org.opentosca.toscana.plugins.kubernetes.model;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.opentosca.toscana.core.testdata.TestCsars;
+import org.opentosca.toscana.core.transformation.Transformation;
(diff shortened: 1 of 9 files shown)


server/src/integration/java/org/opentosca/toscana/plugins/kubernetes/model/KubernetesPushingGopherIT.java
```diff
@@ -0,0 +1,144 @@
...
+            .hostConfig(hostConfig)
+            .image(EXPECTED_TAG).build();
+        String id = client.createContainer(config).id();
+        this.runningContainers.add(id);
+
+        client.startContainer(id);
+
+        Thread.sleep(5000);
+        OkHttpClient httpClient = new OkHttpClient();
+        Request request = new Request.Builder().url("http://127.0.0.1:17717/").get().build();
```
JensMueller2709: Consider to make the port `17717` as a constant. I already saw this port above

c-mueller: Had to make the port dynamic anyways.

PR #577, opened 2018-02-18
Adapt Lamp Models using getEnv

JensMueller2709:
If ok, I will create the new Csars


server/src/test/resources/csars/yaml/valid/task-translator/template.yaml
```diff
@@ -79,20 +79,18 @@ topology_template:
...
-              primary: my_app/create_myphpapp.sh
-              dependencies:
-                - my_app/myphpapp.php
-                - my_app/mysql-credentials.php
-          configure:
-            implementation: my_app/configure_myphpapp.sh
+          start:
             inputs:
-              database_host: { get_attribute: [ db_server, private_address ] }
+              database_host: { get_attribute: [ server, private_address ] }
```
c-mueller: Please revert this to `db_server` since we want the private address of the database compute node!

server/src/test/resources/csars/yaml/valid/lamp-multinode/template.yaml
```diff
@@ -22,20 +22,18 @@ topology_template:
...
-              primary: my_app/create_myphpapp.sh
-              dependencies:
-                - my_app/myphpapp.php
-                - my_app/mysql-credentials.php
-          configure:
-            implementation: my_app/configure_myphpapp.sh
+          start:
             inputs:
-              database_host: { get_attribute: [ db_server, private_address ] }
+              database_host: { get_attribute: [ server, private_address ] }
```
c-mueller: Please revert this to `db_server` otherwise the transformation will fail...

Issue #555, opened 2018-02-15
Aws deploy command does not work with parameters


mmunozba:
**Cause:**
The Plugin adds one too many spaces after --parameter-overrides

mmunozba:
Resolved with #582.

Issue #553, opened 2018-02-15
SSH connection with KeyPair on AWS does not work


mmunozba:
**Cause:**
KeyName is added, but the port isn't open.
**Solution:** 
Open Port 22 on EC2 instances.

mmunozba:
Resolved with #582.

Issue #552, opened 2018-02-15
Delete bucket in cleanup script does not work


mmunozba:
**Cause:**
The command fails because there are still files in the bucket.
**Solution:**
Add `--force` to the command:
https://docs.aws.amazon.com/AmazonS3/latest/dev/delete-or-empty-bucket.html#delete-bucket-awscli

mmunozba:
Resolved with #582.

PR #582, opened 2018-02-22
Fix various AWS bugs

mmunozba:
Fixes
- #552 
- #553 
- #555

### DIFF
--- a/server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
+++ b/server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
@@ -26,7 +26,7 @@ public class CloudFormationFileCreator {
     public static final String CLI_COMMAND_DELETEBUCKET = "aws s3 rb s3://";
     public static final String CLI_PARAM_STACKNAME = "--stack-name ";
     public static final String CLI_PARAM_TEMPLATEFILE = "--template-file ";
-    public static final String CLI_PARAM_PARAMOVERRIDES = "--parameter-overrides ";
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -23,12 +23,13 @@
...
     public static final String CLI_COMMAND_DELETESTACK = "aws cloudformation delete-stack ";
-    public static final String CLI_COMMAND_DELETEBUCKET = "aws s3api delete-bucket --bucket ";
+    public static final String CLI_COMMAND_DELETEBUCKET = "aws s3 rb s3://";
     public static final String CLI_PARAM_STACKNAME = "--stack-name ";
     public static final String CLI_PARAM_TEMPLATEFILE = "--template-file ";
-    public static final String CLI_PARAM_PARAMOVERRIDES = "--parameter-overrides ";
+    public static final String CLI_PARAM_PARAMOVERRIDES = "--parameter-overrides";
     public static final String CLI_PARAM_CAPABILITIES = "--capabilities";
     public static final String CLI_PARAM_BUCKET = "--bucket ";
+    public static final String CLI_PARAM_FORCE = "--force";
```
JensMueller2709: You could consider add a blank in the end of these strings, so you could save additional blanks in your string buildings.

mmunozba: I'm gonna do this in a future PR, all these Strings should be standardised anyways since some of them have a blank and some of them don't.

PR #583, opened 2018-02-22
Support environment variables

mmunozba:


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/handler/EnvironmentHandler.java
```diff
@@ -0,0 +1,121 @@
...
+    private void addSetEnvScriptsToInstances() {
+        logger.debug("Adding setEnv scripts to Instances.");
+        for (Map.Entry<String, Map<String, String>> instanceEnvironment : environmentMap.entrySet()) {
+            String nodeName = instanceEnvironment.getKey();
+            String filePath = SET_ENV + instanceEnvironment.getKey() + ".sh";
+            String cfnSource = getFileURL(cfnModule.getBucketName(), filePath);
+
+            CFNFile cfnFile = new CFNFile(ABSOLUTE_FILE_PATH + filePath)
+                .setSource(cfnSource)
+                .setMode(MODE_500) //TODO Check what mode is needed (read? + execute?)
```
JensMueller2709: still TODO?

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/handler/OperationHandler.java
```diff
@@ -80,7 +96,7 @@ public void handleStart(RootNode node, String computeHostName) {
      @param filePath   path to the artifact
      @param serverName name of the Instance
      */
-    public void addCreate(String filePath, String serverName) throws IOException {
+    public void addCreate(String filePath, String serverName) {
         markUtilFile(filePath);
         CFNFile cfnFile = handleOperationFile(filePath, MODE_500, serverName);
```
JensMueller2709: Is there also a TODO to check which mode is needed or is `MODE_500` the right one?

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/handler/EnvironmentHandler.java
mmunozba: At this point, the modes and owners are probably correct. When we find the time, we'll double check these again until then we'll keep #406 open.

PR #587, opened 2018-02-26
[PROPOSAL] Use MkDocs to generate HTML Documentation

c-mueller:
While trying to get Sphinx running. (Does not work because of some stupid dependency issue)
I created a MkDocs configuration that we could also use to create our documentation. It also supports "Deployment" to readthedocs. (http://www.mkdocs.org/user-guide/deploying-your-docs/)

If you want to take a look a small sample with the current documentation can be seen here:
https://caarl.failwave.de

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,4 @@ app/node
 **/.vscode
 
 !translator.jar
+site/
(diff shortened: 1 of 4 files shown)


nfode:
Looks good.

PR #589, opened 2018-02-26
Execute sql scripts added as artifacts

mklopp:


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/TransformModelNodeVisitor.java
```diff
@@ -171,6 +172,18 @@ public void visit(MysqlDatabase node) {
                 .allocatedStorage(allocatedStorage)
                 .storageType(storageType)
                 .vPCSecurityGroups(cfnModule.fnGetAtt(securityGroupName, "GroupId"));
+            //handle sql artifact
```
mmunozba: If the sql artifact is handled here, we should remove the `TODO` under `visit(MysqlDbms node)`.

server/src/main/java/org/opentosca/toscana/plugins/cloudformation/CloudFormationFileCreator.java
```diff
@@ -171,6 +173,8 @@ private void writeCleanUpScript() throws IOException {
         BashScript cleanupScript = new BashScript(cfnModule.getFileAccess(), FILENAME_CLEANUP);
         // Delete Bucket
         cleanupScript.append(CLI_COMMAND_DELETEBUCKET + cfnModule.getBucketName() + " " + CLI_PARAM_FORCE);
+        // Echo delete Stack
+        cleanupScript.append(COMMAND_ECHO + " " + STRING_DELETESTACK + " " + cfnModule.getStackName());
```
JensMueller2709: The `COMMAND_ECHO` already contains a blank, do you need a additional one?

You could add to the `STRING_DELETESTACK` a blank as well.

PR #590, opened 2018-02-27
Open ports to grant access to WebApp

mklopp:
At some point we lost access to the webapp because the port 80 wouldn't be open


server/src/main/java/org/opentosca/toscana/plugins/cloudformation/visitor/PrepareModelNodeVisitor.java
```diff
@@ -24,8 +26,9 @@
     private static final String AWS_INSTANCE_PRIVATE_IP = "PrivateIp";
     private static final String AWS_INSTANCE_PUBLIC_IP = "PublicIp";
     private static final int minPWLength = 8;
-    private static final String DEFAULT_USER = "root";
-    private static final Integer DEFAULT_PORT = 3306;
+    private static final String DEFAULT_DB_USER = "root";
+    private static final int DEFAULT_DB_PORT = 3306;
+    private static final int DEFAULT_WEBAPP_PORT = 80;
```
c-mueller: Maybe consider renaming "Webapp" to http port

PR #599, opened 2018-02-28
Fix Trailing slash in registry URL

### LINKED ISSUE #513: Registry URL joining mechanism not working properly
While trying to push to a different registry (not dockerhub) i found out that the registry url property currently needs trailing `/` in order to tag the images properly. This is of course not intended and should be fixed

c-mueller:
Fixes #513


c-mueller:
I wont await a Review for this, because this is a very minor fix.

PR #606, opened 2018-03-08
Reformatting Outputs of the CLI

### LINKED ISSUE #605: Fix some CLI Outputs
Fix Outputs of the CLI to improve readability:

- Logs
- System Status
- and other small fixes

orferch:
Fixes #605


cli/src/main/java/org/opentosca/toscana/cli/ApiController.java
```diff
@@ -49,7 +52,6 @@ public String uploadCsar(File file) {
         } catch (IOException e) {
             System.err.println(String.format(SOMETHING_WRONG + Constants.CSAR_UPLOAD_IO_ERROR
                 + Constants.ERROR_PLACEHOLDER, e.getMessage()));
-            e.printStackTrace();
```
nfode: Removing stacktraces which provide important information is nooutpout reformating?

orferch: I removed the Stacktraces because a shorter version of the error is shown right before the stacktrace and the stacktrace is only useful for someone who understands it. 
We can put it back in, or maybe make it a comment and someone who wants the additional information uncomments it.

hnicke: You could show stracktraces when the user calls the cli with the verbose-flag.

cli/src/main/java/org/opentosca/toscana/cli/commands/AbstractApiCall.java
```diff
@@ -7,12 +7,11 @@
...
 public abstract class AbstractApiCall implements Runnable {
 
-    protected ApiController api;
-
-    @CommandLine.Option(names = {"-v", "--verbose"}, description = "Enable Info Level Process Output")
-    private boolean showVerbose;
     @CommandLine.Option(names = {"-m", "--moreverbose"}, description = "Enable Debug Level Process Output")
-    private boolean showMVerbose;
+    public boolean showMVerbose;
+    @CommandLine.Option(names = {"-v", "--verbose"}, description = "Enable Info Level Process Output")
```
nfode: What is printed if I use the verbose flag?

cli/src/main/java/org/opentosca/toscana/cli/ApiController.java
```diff
@@ -49,7 +53,7 @@ public String uploadCsar(File file) {
...
```
nfode: This line also needs to be changed

cli/src/main/java/org/opentosca/toscana/cli/commands/AbstractApiCall.java
orferch: Verbose -v gives you a "shorter" output then -m. See here:

verbose

PR #611, opened 2018-03-17
Improve Javadoc of the KubernetesPlugin

c-mueller:


c-mueller:
I did get rid of some of them, at points where no rewriting of the sentence is required.

I will merge this soon, because the lists of tasks that have to be done until saturday is still quite long.

PR #613, opened 2018-03-21
Developer documentation for EffectiveModel

hnicke:
Writing developer documentation for the EffectiveModel and the ServiceGraph.

It's not 100% complete but the essential part is done.


hnicke:
Please take special care when reviewing ADR n17. Is it comprehensible, even for an outsider?

hnicke:
Will merge this now without more feedback. Need to put all the PR's together in order to continue writing docs.

PR #618, opened 2018-03-24
Add user documentation about TOSCA support

hnicke:
This PR adds user documentation about TOSCA support and limitations, targeted specification, TOSCAna types etc.

Currently, I think that's all I have to say in the user documentation. Did I miss something?


docs/adr/0016-not-supporting-type-definitions.md
```diff
@@ -0,0 +1,29 @@
+# Do not support type definitions in TOSCA service template
+
+**User Story:** 
+
+Should *TOSCAna* support node type definitions in the TOSCA service template?
+*[decision drivers | forces]* <!-- optional -->
```
c-mueller: I would remove this line, since the "question" isnt answered

hnicke: Do you mean line 5 or line 6? Line 6 must get removed definitely.

c-mueller: Line 6

PR #631, opened 2018-03-29
All in one download script

nfode:
blocked by #628


app/src/app/components/transformation-view/transformation-view.component.ts
```diff
@@ -117,6 +117,25 @@ export class TransformationViewComponent implements OnInit, OnDestroy {
...
+        const folder = this.csarId + '-' + this.platform;
+        let result = '#!/bin/bash\n';
+        result += 'timestamp=$(date +%s)\n';
+        result += 'folder=' + folder + '-$timestamp\n';
+        result += 'mkdir $folder\n';
+        result += 'cd $folder\n';
+        result += 'curl ' + this.url + ' --output $folder.csar\n';
+        result += 'unzip $folder.csar -d unzipped\n';
+        result += 'bash unzipped/output/scripts/deploy.sh\n';
+        result += 'cd ../\n';
```
hnicke: Automatic cleanup after script execution...hm. Not sure if this is user-friendly. What if user aborts the execution of the deploy.sh and wants to execute it (e.g., with different parameters?) 
My idea is to prompt the user for cleanup: `cleanup [y/N]?`
Maybe also prompt for execution of deploy.sh. so user has the chance to modify the script.

nfode: This might be a good idea going to add this tomorrow

app/src/app/components/transformation-view/transformation-view.component.ts
```diff
@@ -128,6 +128,14 @@ export class TransformationViewComponent implements OnInit, OnDestroy {
         result += 'unzip $folder.csar -d unzipped\n';
         result += 'bash unzipped/output/scripts/deploy.sh\n';
         result += 'cd ../\n';
+        result += 'printf "cleanup [y/N]: "\n' +
+            'read input\n' +
+            '\n' +
+            'if [ "$input" != "y" ] && [ "$input" != "yes" ] && [ "$input" != "Y" ]\n' +
```
hnicke: Don't be afraid of using regexp: `if [[ ! "$input" =~ ^(y|yes|Y)$ ]]; then`
Regex is awesome!

app/src/app/components/transformation-view/transformation-view.component.ts
```diff
@@ -128,6 +128,14 @@ export class TransformationViewComponent implements OnInit, OnDestroy {
...
+            'then\n' +
+            '  echo "Cleanup skipped!"\n' +
+            '  exit 1\n' +
```
hnicke: Why would you exit with 1? This is not an error.

